PR #7, opened 2022-12-07
Add Paganin filter for the pipeline

namannimmo10:


tests/test_filtering.py
```diff
@@ -33,6 +35,24 @@ def test_fresnel_filter():
     pattern = 'SINOGRAM'
     filtered_data = fresnel_filter(data, pattern, ratio)
     for _ in range(5):
-        assert_allclose(cp.mean(filtered_data), 806.74347, rtol=1e-06)
+        assert_allclose(cp.mean(filtered_data), 806.74347, rtol=eps)
         assert_allclose(cp.max(filtered_data), 1063.7007)
         assert_allclose(cp.min(filtered_data), 87.91508)
+
+    #--- testing the Paganin filter on tomo_standard ---#
```
yousefmoazzam, line 42: I think for the sake of separation and clarity of the test results, the test of the Paganin filter should be in a separate function to the test of the Fresnel filter, something like `test_paganin_filter()`? Otherwise, if it were the case that `test_fresnel_filter()` is failing, the first instinct is that something is wrong with the Fresnel filter, but it could actually be the Paganin filter that's failing?

I know that (at the moment) they both use the same input data so it's convenient to load the data in only once and run it through both filters in a single function, but since these are tests it feels nice to have them separate. I'm sure some combination of setup/teardown methods could be used if the duplication of the data loading in the two tests is bothersome, I think [fixtures](https://docs.pytest.org/en/7.1.x/explanation/fixtures.html) are the way pytest does this kind of thing?

src/httomolib/filtering.py
```diff
@@ -92,3 +94,96 @@ def _make_window(height, width, ratio, pattern):
...
+
+    increment : optional, float
+        Increment all values by this amount before taking the log.
+
+    Returns
+    -------
+    cp.ndarray
+        The stack of filtered projections.
+    """
+    if data.ndim == 2:
```
yousefmoazzam, line 141: I guess after any 2D data has been promoted to 3D, we can safely consider all data that is _not_ 3D to be invalid input? In which case, raising a `ValueError` like you had suggested after this `if` statement, something like:
```
if data.ndim !=3:
    raise ValueError(f"Invalid number of dimensions in data: {data.ndim}, please provide a stack of 2D projections.")
```
could be useful.

yousefmoazzam (changes requested):
Thanks, the test for the Paganin filter passes for me! A few suggestions, let me know what you think.

tests/test_filtering.py
namannimmo10: I can refactor the tests later, I think it's ok to have `data` as a global variable for now.  We can also import `data` from a file like `tests/utils.py`.

namannimmo10:
Thank you much for the suggestions, please let me know if it looks good to merge now.

yousefmoazzam:
Cool, all looks good to me!

PR #11, opened 2022-12-14
Set up the basic CI infrastructure

namannimmo10:


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
```
aktech: If you want cirun to spinup runners for this job, you need to change this line to match the labels defined in your `.cirun.yml`:

```suggestion
    runs-on: cirun-gpu-runner
```

namannimmo10: Ahh, thank you, Amit. Just playing with this for the first time!

aktech:
Since you're using an image from the AWS Marketplace, you'd need to accept their terms, see here:
https://github.com/DiamondLightSource/httomolib/pull/11/checks?check_run_id=10980273427

namannimmo10:
Where do I accept the terms?

aktech:
Checkout the link above, there should be terms link in the error:


Screenshot 2023-01-30 at 4 32 31 pm

```
{
    "error": "An error occurred (OptInRequired) when calling the RunInstances operation: In order to use this AWS Marketplace product you need to accept terms and subscribe. To do so please visit https://aws.amazon.com/marketplace/pp?sku=46kqvw7d9nbo0v2bfgi8z26gb"
}
```

This link basically: https://aws.amazon.com/marketplace/pp?sku=46kqvw7d9nbo0v2bfgi8z26gb

namannimmo10:
Hi Amit, for me it says, 

>  This version has been removed and is no longer available to new customers.

Do you suggest we use some other GPU (which is free)?

aktech:
You can try this one: https://aws.amazon.com/marketplace/pp/prodview-7ikjtg3um26wq?sr=0-9&ref_=beagle&applicationId=AWSMPContessa

This is the latest version of the previous one. Also, note that both of these are free (previous one too). The cost you see on the right is the cost of the instance not the image (which you'd pay anyway irrespective of which image you use).

PR #16, opened 2022-12-19
data renamed, dependencies on httomo dropped, tests rearranged

### LINKED ISSUE #15: create test data that does not require HTTomo loaders
* Keep all I/O for HDF5 in HTTomo and do not use it in httomolib. As a library of methods, lets keep things very simple regarding the data reads. I suggest we create a compressed `npz` file using this [function](https://numpy.org/doc/stable/reference/generated/numpy.savez_compressed.html#numpy.savez_compressed) from hdf5 files.
* Modify unittests accordingly

dkazanc:
Fixes #15 
* data renamed, dependencies on httomo dropped, tests rearranged
* most of dependencies dropped for httomolib as redundant. We want to make httomolib with only required dependencies, the ones from HTTomo should be removed.


namannimmo10:
Amazing speed-up while running the tests--thank you!

README.rst
```diff
@@ -4,20 +4,19 @@ httomolib
 Setup
 =====
 * Clone the repository from GitHub using :code:`git clone git@github.com:DiamondLightSource/httomolib.git`
-* Install dependencies from the environment file :code:`conda env create httomolib --file conda/environment.yml` (SLOW)
-* Alternatively you can install from the existing explicit file :code:`conda create --name httomolib --file conda/explicit.txt`
+* Install dependencies from the environment file :code:`conda env create httomolib --file conda/environment.yml`
```
namannimmo10, line 7: This is the right way to do it but, it is also _really_ slow. What do you think?

namannimmo10:
We could possibly update the [examples](https://github.com/DiamondLightSource/httomolib/blob/master/examples/fresnel-filter.py), too.

README.rst
namannimmo10: Maybe we can use [Mamba](https://pythonspeed.com/articles/faster-conda-install/) to speed-up the conda installs.

dkazanc:
ah yes, go ahead and update examples if you like. Using mamba it's fine to accelerate things. We will add explicit installation option as we add more dependencies in future.

tests/test_filtering.py
```diff
@@ -1,23 +1,15 @@
...
-dimension = 1
-preview = [None, None, None]
-pad = 0
-eps = 1e-6
-host_data = \
-    standard_tomo(in_file, data_key, image_key, dimension, preview, pad, comm)[0]
+in_file = 'data/tomo_standard.npz'
+datafile = np.load(in_file) #keys: data, flats, darks, angles, angles_total, detector_y, detector_x
+host_data = datafile['data']
 data = cp.array(host_data)
```
namannimmo10, line 11: Why don't we simply use `cp.load`?

namannimmo10: It should work.

namannimmo10: But we will eventually have CPU methods too, so might be worth keeping `host_data`. Scratch my comment, please!

dkazanc: sure, either is fine.

namannimmo10 (approved):
Looks good for now, thanks.

PR #24, opened 2023-01-06
Add a basic CPU vs GPU test for the whole pipeline

namannimmo10:


namannimmo10:
This is based on the pipeline order that Yousef gave me sometime before,

1. Raw data
2. Normalisation
3. Remove stripes
4. Fresnel/Paganin filter
5. Centering
6. Reconstruction

src/httomolib/prep/normalize.py
```diff
@@ -121,30 +120,45 @@ def normalize_cupy(
...
-        3D flat field data as a CuPy array.
+        2D or 3D flat field data as a CuPy array.
     darks : ndarray
-        3D dark field data as a CuPy array.
+        2D or 3D dark field data as a CuPy array.
     cutoff : float, optional
         Permitted maximum value for the normalised data.
     minus_log : bool, optional
-        Apply negative log to the normalised
+        Apply negative log to the normalised data
```
namannimmo10, line 131: If this implementation replicates the one from TomoPy, this should have the same parameters, I believe... so we need `averaging='mean'` or `'median'`: https://github.com/tomopy/tomopy/blob/c236a2969074f5fc70189fb5545f0a165924f916/source/tomopy/prep/normalize.py#L122

yousefmoazzam: This makes sense if we are indeed replicating the TomoPy implementation.

However, perhaps things have moved on since I last worked on the normalisation, but I recall there being an idea that we're unsure if the TomoPy implementation should be used as the "reference" for the dark flat field correction method; something about TomoPy handling negative values or zeros weirdly when taking the negative log, @dkazanc does that sound familiar or am I speaking nonsense?

namannimmo10:
a) Is it better to add `minus_log` as a separate function just like [tomopy](https://github.com/tomopy/tomopy/blob/c236a2969074f5fc70189fb5545f0a165924f916/doc/demo/gridrec.py#L78) does? 
b) The method `normalize_cupy` replicates the one from tomopy, so should I address https://github.com/DiamondLightSource/httomolib/pull/24#discussion_r1072325328? 

Let me know and I will complete this PR.

src/httomolib/prep/normalize.py
dkazanc: that's correct @yousefmoazzam, I think we can step away a bit from TomoPy implementation to deal with nans, infs and negative values. In general, I think it is OK to have our own implementation as long as we know what we're doing and the GPU implementation repeats the CPU one. I think will add my implementation of the normalisation step in httomolib. Also there is not much sense to have `minus_log` as a separate function but a parameter @namannimmo10 . In fact, we should reduce the granularity of methods where we can.

PR #25, opened 2023-01-06
Adds the preprocessing pipeline

dkazanc:


examples/pipeline_preproc.py
```diff
@@ -0,0 +1,68 @@
...
+host_darks = datafile['darks']
+
+# Apply CuPy implementation of normalisation to data
+data = cp.asarray(host_data)
+flats = cp.asarray(host_flats)
+darks = cp.asarray(host_darks)
+
+elapsed_time_total = 0.0
+# ---------------------------------------------------------------#
+#TODO: dezinger should be added here
```
namannimmo10, line 31: Why do we add the filter over here? Don't we normalize the data first?

dkazanc: because we usually process the raw data plus darks and flats. It does a better job than after normalisation

namannimmo10:
This is what I get: 
```console
$ python examples/pipeline_preproc.py 
Traceback (most recent call last):
  File "/home/qfc83269/Applications/oss/httomolib/examples/pipeline_preproc.py", line 19, in <module>
    datafile = np.load(in_file)
  File "/dls/science/users/qfc83269/condaenvs/httomolib/lib/python3.9/site-packages/numpy/lib/npyio.py", line 405, in load
    fid = stack.enter_context(open(os_fspath(file), "rb"))
FileNotFoundError: [Errno 2] No such file or directory: '../tests/test_data/tomo_standard.npz'

```

PR #26, opened 2023-01-09
Improve httomolib.prep.phase

namannimmo10:


src/httomolib/prep/phase.py
```diff
@@ -32,13 +32,19 @@
 
 
 # CuPy implementation of Fresnel filter ported from Savu
-def fresnel_filter(mat: cp.ndarray, pattern: str, ratio: float,
-                   apply_log: bool = True):
-    """Apply Fresnel filter.
+def fresnel_filter(
+    data: cp.ndarray,
+    pattern: str,
+    ratio: float,
```
namannimmo10, line 38: @dkazanc : Can `pattern='PROJECTION'` be the default here?

PR #30, opened 2023-01-13
Image saver changes

dkazanc:
@namannimmo10 could you please add a test to the function. thanks!

### DIFF
--- a/src/httomolib/misc/images.py
+++ b/src/httomolib/misc/images.py
@@ -147,7 +147,7 @@ def _save_single_img(array2d, glob_stats, bits, jpeg_quality, path_to_out_file):
             array2d,
             in_range=(data_min, data_max),
             out_range=(data_min, data_max)
-        ).astype(np.float32)
+        ).astype(np.uint32)
 
     img = Image.fromarray(array2d)
     img.save(path_to_out_file, quality=jpeg_quality)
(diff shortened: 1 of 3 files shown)


namannimmo10:
Sure, I will add a test. 

You need to `git rebase` with the `master` branch, and then add your new commits on top of that.

src/httomolib/misc/images.py
```diff
@@ -0,0 +1,143 @@
...
+# Created By  : Daniil Kazantsev <scientificsoftware@diamond.ac.uk>
+# Created Date: 27/October/2022
+# version ='0.1'
+# ---------------------------------------------------------------------------
+""" Module for loading/saving images """
+
+import os
+
+import numpy as np
+import skimage.exposure as exposure
```
namannimmo10:
```suggestion
from skimage import exposure
```

src/httomolib/misc/images.py
```diff
@@ -0,0 +1,153 @@
...
+            in_range=(data_min, data_max),
+            out_range=(0, 65535)
+        ).astype(np.uint16)
+
+    else:
+        array2d = exposure.rescale_intensity(
+            array2d,
+            in_range=(data_min, data_max),
+            out_range=(data_min, data_max)
+        ).astype(np.float32)
```
namannimmo10: Why is this different for `bits=32`?
```suggestion
        ).astype(np.uint32)
```

namannimmo10: Actually we need to make this change otherwise the test fails for `bits=32`

PR #38, opened 2023-01-18
GPU reconstruction and other changes

dkazanc:


src/httomolib/prep/normalize.py
```diff
@@ -48,59 +51,79 @@ def normalize_raw_cuda(
...
-    params = (data, flat0, dark0, out, float32(1e-6),
-              float32(cutoff), data.shape[1], data.shape[2])
+    take_log_param = 0
+    if minus_log:
+        take_log_param = 1
+    nonnegativity_param = 0
+    if nonnegativity:
+        nonnegativity_param = 1
+    params = (data, flat0, dark0, out, float32(cutoff),
+              take_log_param, nonnegativity_param, data.shape[1], data.shape[2])
```
namannimmo10: CUDA toolkit supports `bool`.
```suggestion
    params = (data, flat0, dark0, out, float32(cutoff),
              minus_log, nonnegativity, data.shape[1], data.shape[2])
```

src/httomolib/prep/normalize.py
```diff
@@ -48,59 +51,79 @@ def normalize_raw_cuda(
...
+                    tmp = cutoff;
+                    }      
+                    if (take_log_param == 1)
+                    {
+                        tmp = -logf(tmp);
+                    }
+                    if (nonnegativity_param == 1)
+                    {
+                        if (tmp < 0.0f) {tmp = 0.0f;}
+                    }
```
namannimmo10: I guess we need to handle `NaN` to `0.` conversion here as well.

src/httomolib/prep/normalize.py
dkazanc: ah great, wasn't sure about bool.

src/httomolib/prep/normalize.py
dkazanc: good point, in fact I don't know what CUDA will produce with the logf(0.0), nothing good probably. Will check it now

src/httomolib/prep/normalize.py
namannimmo10: Data types: https://docs.cupy.dev/en/stable/user_guide/kernel.html#kernel-arguments

src/httomolib/prep/normalize.py
```diff
@@ -34,7 +34,10 @@ def normalize_raw_cuda(
     data: ndarray,
     flats: ndarray,
     darks: ndarray,
-    cutoff: float = 10.
+    gpu_id : int = 0,
+    cutoff: float = 10.0,
+    minus_log: bool = False,
+    nonnegativity: bool = False
```
namannimmo10:
```suggestion
    absolute: bool = False
```

namannimmo10:
I think we are making the code complicated for us with _many_ CPU - GPU transfers.
I recommend reading the [Zen of Python](https://peps.python.org/pep-0020/#the-zen-of-python) :)

src/httomolib/recon/algorithm.py
```diff
@@ -0,0 +1,147 @@
...
+    rn2 = np.sin(a*w/2.0)
+    rd = (a*w)/2.0
+    rd_c = np.zeros([1,DetectorsLengthH])
+    rd_c[0,:] = rd
+    r = rn1*(np.dot(rn2, np.linalg.pinv(rd_c)))**2
+    multiplier = (1.0/projectionsNum)
+    f = scipy.fftpack.fftshift(r)
+    f_2d = np.zeros((DetectorsLengthV,DetectorsLengthH), dtype='float32') # 2d filter
+    f_2d[0::,:] = f
+    filter_gpu = cp.asarray(f_2d)
```
namannimmo10: If you know that the function argument `projection3D` is a cupy array, then let's just simply use cupy functions here, instead of numpy -> cupy.

dkazanc: agree, will correct now

namannimmo10: We don't need scipy here: 
```python

In [3]: freqs = cp.fft.fftfreq(10, 0.1)

In [4]: freqs
Out[4]: array([ 0.,  1.,  2.,  3.,  4., -5., -4., -3., -2., -1.])

In [5]: cp.fft.fftshift(freqs)
Out[5]: array([-5., -4., -3., -2., -1.,  0.,  1.,  2.,  3.,  4.])

In [6]: freqs_np = np.fft.fftfreq(10, 0.1)

In [7]: import scipy

In [8]: scipy.fftpack.fftshift(freqs_np)
Out[8]: array([-5., -4., -3., -2., -1.,  0.,  1.,  2.,  3.,  4.])
```

src/httomolib/prep/normalize.py
dkazanc: it is a historical term a bit...

src/httomolib/recon/algorithm.py
dkazanc: yep, I got rid of scipy.

src/httomolib/recon/algorithm.py
```diff
@@ -0,0 +1,192 @@
...
+        A GPU device index to perform operation on.
+
+    Returns
+    -------
+    np.ndarray
+        The reconstructed volume.
+    """
+    from tomopy import astra, recon
+    
+    reconstruction = recon(cp.asnumpy(data),
```
namannimmo10, line 185:
```suggestion
    reconstruction = recon(data,
```

dkazanc: in this case it needs a numpy array but data is a cupy one so it will result in the error...

namannimmo10: Well, the documentation said otherwise: data is a numpy array.

yousefmoazzam: @namannimmo10 The tomopy recon function certainly takes a numpy array, but I think what @dkazanc  means is that the array that the `reconstruct_tomopy()` is given will be a cupy array (from previous methods in the pre-processing stage), so it first needs to be converted from a cupy array to a numpy array before being passed to a tomopy recon algorithm, is that right @dkazanc ?

If so, the type signature and docstring  of the `reconstruct_tomopy()` function needs to be updated to say it takes in a `cp.ndarray` rather than `np.ndarray`, but I think this code is correct otherwise?

dkazanc: That's correct. You can have a look at the wrapper function in httomo where I convert to cupy array explicitly. I think this is where the checks should be of what data type is being passed based on the method selection. We better write things in httomolib as httomo doesn't exist. So if function works with numpy only it should be stated and numpy should be given. Does it makes sense?

src/httomolib/prep/normalize.py
```diff
@@ -65,42 +76,48 @@ def normalize_raw_cuda(
...
+                    tmp = cutoff;
+                    }      
+                    if (take_log_param == 1)
+                    {
+                        tmp = -logf(tmp);
+                    }
+                    if (nonnegativity_param == 1)
+                    {
+                        if (tmp < 0.0f) {tmp = 0.0f;}
+                    }
```
namannimmo10:
```suggestion
                    if (minus_log)
                    {
                        tmp = -logf(tmp);
                    }
                    if (nonnegativity)
                    {
                        if (tmp < 0.0f) {tmp = 0.0f;}
                    }
```

src/httomolib/prep/normalize.py
```diff
@@ -65,42 +76,48 @@ def normalize_raw_cuda(
         """extern "C" __global__ void normalize(const unsigned short* data,
            const float* flat,
            const float* dark,
-           float* out, float eps, float cutoff, int A, int B)
+           float* out, float cutoff, int take_log_param, int nonnegativity_param, int A, int B)
```
namannimmo10:
```suggestion
           float* out, float cutoff, bool minus_log, bool nonnegativity, int A, int B)
```

src/httomolib/prep/normalize.py
```diff
@@ -48,15 +51,23 @@ def normalize_raw_cuda(
...
+        Remove negative values in the normalised data.          
 
     Returns
     -------
     cp.ndarray
         Normalised 3D tomographic data as a CuPy array.
     """
 
+    cp.cuda.Device(gpu_id).use()
+
```
namannimmo10:
```suggestion

    if data.ndim != 3:
        raise ValueError("Input data must be a 3D stack of projections")
```
Similar to `normalize_cupy`.

src/httomolib/misc/corr.py
```diff
@@ -22,13 +22,128 @@
...
+        Median filtered 3D CuPy array either float32 or uint16 data type.
+
+    Raises
+    ------
+    ValueError
+        If the input array is not three dimensional.
+
+    """
+    input_type = data.dtype
+    if (input_type != 'float32') and (input_type != 'uint16'):
```
namannimmo10:
```suggestion
    if input_type not in ["float32", "uint16"]:
```

src/httomolib/prep/normalize.py
```diff
@@ -125,40 +147,53 @@ def normalize_cupy(
...
+    if flats.ndim not in (2, 3):
+        raise ValueError("Input flats must be 2D or 3D data only")
+
+    if darks.ndim not in (2, 3):
+        raise ValueError("Input darks must be 2D or 3D data only")
 
     # replicates tomopy implementation
     denom = (flats - darks)
-    denom[denom < eps] = eps
+    denom[denom < 1.0] = 1.0
```
namannimmo10: Now, the issue I see here is that we have made this implementation different from TomoPy's. That is why it makes less sense to have a CPU - GPU test: `tests/test_pipeline.py::test_cpu_vs_gpu` fails because we get different values.

dkazanc: my mistake here, will fix it now.

src/httomolib/prep/normalize.py
```diff
@@ -125,40 +49,53 @@ def normalize_cupy(
...
+        raise ValueError("Input flats must be 2D or 3D data only")
+
+    if darks.ndim not in (2, 3):
+        raise ValueError("Input darks must be 2D or 3D data only")
 
     # replicates tomopy implementation
+    lowval_threshold = cp.float32(1e-6)
     denom = (flats - darks)
-    denom[denom < eps] = eps
+    denom[denom < lowval_threshold] = lowval_threshold
```
namannimmo10:
```suggestion
    denom[denom < lowval_threshold] = lowval_threshold
```

PR #39, opened 2023-01-18
Add test for distortion correction from tomopy

namannimmo10:


tests/test_prep/test_alignment.py
```diff
@@ -34,6 +44,13 @@ def test_correct_distortion():
         assert_allclose(cp.mean(corrected_data), 95.51871109008789)
         assert cp.max(corrected_data) == 228
 
+    #: test the implementation from tomopy
+    corrected_data = distortion_correction_proj_discorpy_cupy(
+        im, distortion_coeffs_path, preview)
+    for _ in range(5):
+        assert_allclose(cp.mean(corrected_data), 87.76617813110352)
```
namannimmo10, line 51: Looks like the values from Savu and TomoPy/Discorpy don't match... is it fair?
@dkazanc @yousefmoazzam

yousefmoazzam: A few comments on my end:
- I did some calculations locally during the porting, the residual between the Savu output and the CuPy port of the Discorpy implementation is indeed non-zero but reasonably small: I took their difference and saw mostly zeroes with some 1's and 2's dotted around in the array, then summed the rows and columns, which resulted in the value `238`.
- When going through the Savu and Discorpy implementations to do the port, it first appeared that they were somewhat different. However, if you compare the two ports side-by-side there's actually not much difference, it's mostly that the Savu implementation has a few extra things for checking that some values aren't nonsense. Ie, they seem to have similar "approaches".

My naive opinion is that:
- the Savu implementation https://github.com/DiamondLightSource/Savu/blob/master/savu/plugins/corrections/distortion_correction.py#L92-L172
- the TomoPy implementation https://github.com/tomopy/tomopy/blob/c236a2969074f5fc70189fb5545f0a165924f916/source/tomopy/prep/alignment.py#L912-L947
- the "backward model" implementation in Discorpy https://github.com/DiamondLightSource/discorpy/blob/67743842b60bf5dd45b21b8460e369d4a5e94d67/discorpy/post/postprocessing.py#L111-L148

are all implementing the same method, the so-called "backward model" that is described in the paper in section 2.3.3 https://opg.optica.org/oe/fulltext.cfm?uri=oe-23-25-32859&id=333610.

If this is the case, then their "approaches" are the same, which means that they shouldn't be producing numerically different results. Given that, it certainly seems possible that my ports introduced some small error into the implementations.

So please do check that things look ok to you @dkazanc as far as how the different results look! Bear in mind that the distortion coefficients in https://github.com/DiamondLightSource/httomolib/blob/master/tests/test_data/distortion-correction/distortion-coeffs.txt were calculated for the dot pattern image only https://github.com/DiamondLightSource/httomolib/blob/master/tests/test_data/distortion-correction/dot_pattern_03.tif, so I don't know if the application of the different implementations to the other test images using that particular text file will produce anything meaningful to the eye.

dkazanc: I think we could stick to the ported version and see how it behaves for the actual data. If it satisfies expectations then there is no need to investigate this further. If, however, there is a problem then I guess we need to double check if the implementation is actually follows equations given in the paper?

Issue #22, opened 2023-01-06
Tomopy methods to port for acceleration


dkazanc:
We need to port:

1.  _remove_all_stripe_, _remove_large_stripe_ and _remove_dead_stripe_ from [here](https://github.com/tomopy/tomopy/blob/master/source/tomopy/prep/stripe.py)
2. retrieve_phase from [here](https://github.com/tomopy/tomopy/blob/master/source/tomopy/prep/phase.py). This is so-called [Paganin filter](https://github.com/DiamondLightSource/httomolib/blob/34c01ffbdf2f53eea5dc17aecd3e4b1051b1cbc8/src/httomolib/prep/phase.py#L128) but we need to make it consistent with TomoPy implementation
3. Distortion correction as in [here](https://github.com/tomopy/tomopy/blob/1cb9411f6c17ca3e6eba72a2e7b67ed4f0e77498/source/tomopy/prep/alignment.py#L912)

yousefmoazzam:
Re: item 3 in the list.

An alternative distortion correction implementation has been added in 050ded9.

It is based on the implementation in Discorpy https://github.com/DiamondLightSource/discorpy/blob/67743842b60bf5dd45b21b8460e369d4a5e94d67/discorpy/post/postprocessing.py#L111-L148 rather than TomoPy https://github.com/tomopy/tomopy/blob/c236a2969074f5fc70189fb5545f0a165924f916/source/tomopy/prep/alignment.py#L950-L981.

This is because they are actually almost identical to each another, except that the Discorpy version is slightly more flexible by allowing users to pass the `order`  and `mode` parameters through to the `map_coordinates()` function from NumPy https://github.com/DiamondLightSource/discorpy/blob/67743842b60bf5dd45b21b8460e369d4a5e94d67/discorpy/post/postprocessing.py#L147.

yousefmoazzam:
It's not urgently needed to be ready for the GPU contractor's start date, but as a note, it was also previously discussed that the methods relating to 360 sinograms would be useful to make GPU ports of sometime, namely:
- Conversion of a 360 degree sinogram to a 180 degree sinogram (from TomoPy): https://github.com/tomopy/tomopy/blob/master/source/tomopy/misc/morph.py#L314-L356
- Finding the CoR of a 360 sinogram (from algotom): https://github.com/algotom/algotom/blob/96ec59e4ae49d937ae26be2e63aa6d0ec55da3a6/algotom/prep/calculation.py#L599-L649

PR #40, opened 2023-01-19
Port more stripe removal methods from TomoPy to CuPy

yousefmoazzam:
This PR will be providing CuPy ports of the methods mentioned in item 1 from #22.

### DIFF
--- a/src/httomolib/prep/_rgi.py
+++ b/src/httomolib/prep/_rgi.py
@@ -1,0 +1,536 @@
+# Code taken from
+# https://github.com/cupy/cupy/blob/1a9c91411fe6297ea81c6ad103cd8211134f6446/cupyx/scipy/interpolate/_rgi.py
+# since the `interpn()` function is only available in CuPy v12.0.0b3 pre-release
+# https://github.com/cupy/cupy/releases/tag/v12.0.0b3
+__all__ = ['RegularGridInterpolator', 'interpn']
+
+import itertools
+import cupy as cp
+
+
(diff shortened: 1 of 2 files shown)


yousefmoazzam:
@dkazanc @namannimmo10 Heads-up, the port of dead stripe removal involves something that people may find a bit dodgy, please read the commit message for e18feaf and let me know if you think an alternative should be done instead!

dkazanc:
@yousefmoazzam I think it is fine to use the ported function for now but we need to make a mark that it should be replaced with the 12.0 rlease of CuPy?

yousefmoazzam:
@dkazanc Yep, that makes sense. I've put code comments in the method https://github.com/DiamondLightSource/httomolib/blob/0f4e060b4510efd42450aa199cf4217475d21896/src/httomolib/prep/stripe.py#L247-L250 and at the top of the copied file https://github.com/DiamondLightSource/httomolib/blob/0f4e060b4510efd42450aa199cf4217475d21896/src/httomolib/prep/_rgi.py#L1-L4 but if there is clearer/more obvious way to mark it then feel free to do so.

dkazanc:
@yousefmoazzam I guess we have now a few hidden functions in the `remove_all_stripe` module ported from TomoCuPy. We can probably turn them into separate methods, which is a different task. We probably then close this PR and make a separate issue on extracting sub-methods from  `remove_all_stripe` into separate exposed modules?

yousefmoazzam:
Yeah, this is pretty outdated, so I'm happy to close this in favour of what you've suggested.

PR #49, opened 2023-01-31
Update project organisation with pyproject.toml

team-gpu:
Reorganises the project structure for a modern Python project, using only `pyproject.toml` and specifying dependencies correctly. In particular:

- all `setup.cfg` and `setup.py` settings go into `pyproject.toml`
- Fills in missing bits and modifies from poetry settings to standard Python `[project]` section in `pyproject.toml`
- Moves from `src/httomolib` to flat structure, with package at the top-level
- also specifies development dependencies (required to run the tests)


conda/environment.yml
```diff
@@ -8,10 +8,16 @@ dependencies:
   - conda-forge::cupy
   - conda-forge::numpy
   - conda-forge::tomopy
+  - conda-forge::scipy
   - conda-forge::python
   - conda-forge::pytest
   - conda-forge::pytest-xdist
+  - conda-forge::pytest-cov
+  - conda-forge::flake8
+  - conda-forge::Flak8-pyproject
```
namannimmo10:
```suggestion
  - conda-forge::Flake8-pyproject
```

namannimmo10: BTW, it looks like this is not under conda-forge: https://conda-forge.org/feedstock-outputs/

yousefmoazzam: Maybe a typo, this looks like it could be it: https://github.com/conda-forge/pyproject-flake8-feedstock and https://anaconda.org/conda-forge/pyproject-flake8?

conda/environment.yml
```diff
@@ -8,10 +8,16 @@ dependencies:
   - conda-forge::cupy
   - conda-forge::numpy
   - conda-forge::tomopy
+  - conda-forge::scipy
   - conda-forge::python
   - conda-forge::pytest
   - conda-forge::pytest-xdist
+  - conda-forge::pytest-cov
```
namannimmo10, line 15: Actually, we can be more specific with the package versions we need -- that will hopefully speed up the conda installation. Worth giving mamba a try as well in the future.

conda/environment.yml
team-gpu: Ah I see. This is the project I meant: https://pypi.org/project/pyproject-flake8/ - it's a typo, fixing it now

team-gpu: (I had this originally from pip, but both packages to the same thing: https://pypi.org/project/Flake8-pyproject/)

conda/environment.yml
team-gpu: Typically for dev-only packages, we don't need to be that specific and latest version will do. Unless we see some breaks. That is, for pytest and related, versioning is not that important. But for numpy/cupy/scipy etc, it's worth being more specific if you're using features that break in older versions, etc.

namannimmo10 (approved):
looks good, thanks much!

PR #50, opened 2023-02-01
Add normalisation raw CUDA kernel

yousefmoazzam:


team-gpu:
You should try to calculate the mean separately, using cupy's mean function, and then use an `Elementwise`  kernel for the rest.

yousefmoazzam:
After seeing you walk us through how you rewrote the Paganin filter using `ElementWise` kernels, I was curious to try out exactly that!

Do you think it would be interesting to compare how this `RawKernel` performs in comparison to a version using `ElementWise` kernel? One thing I was curious about was how the thread-block and grid dimensions are determined when using `ElementWise` kernel, since with `RawKernel` I have chosen a particular (naive) way to set them: https://github.com/DiamondLightSource/httomolib/blob/ccdb0b901ee5ae9cb874cf5bc1334f8774fe511e/httomolib/prep/normalize.py#L76-L88

which could possibly affect the "generality" of the kernel (ie, for different height/width projection images it would be launched with a specific value for the block/grid dims based on the arithmetic above, and if this could affect how it performs)?

For simple processing where the indexing involved in the raw CUDA code isn't too bad, I'm kinda curious to see if `RawKernel` has any performance benefit compared to `ElementWise`, do you have any thoughts on this? Or would it be a silly/pointless thing to check?

team-gpu:
I think if the Elementwise kernel can do the job, you'll have a hard time being faster with a RawKernel, unless you're doing some non-standard operations. Kernel performance is mostly about access patterns and data organisation in memory - and cupy has that figured out depending on the arrays you feed to the call.

Usually a RawKernel makes sense if you're doing something different than strictly elementwise operations.

PR #51, opened 2023-02-01
Test refactorings

team-gpu:
### This PR refactors tests as follows:

- using pytest fixtures to avoid copy/pasting all the data loading
- split up tests that tested several things in a row into standalone tests, to avoid dependence on data and cross-effects and for better results reporting
- parametrize tests where it makes sense
- do assertions / comparisons on the CPU with numpy
- remove repeated assertions of the exact same data - there's no randomness with numpy's assert functions

### Things to note: 

There are 2 TODO comments, showing where a previous function modifies the data in-place, so had to call it again in the split-up test to match the assertions as they were. The reference values for these should ideally be re-based to the single call only.

Further, `numpy` calculates means in a different order to `cupy`, due to the parallel reductions on the GPU. This meant I had to increase tolerances in 2 functions a tiny bit to match the reference - please check this carefully.

### DIFF
--- a/tests/test_recon/test_recon.py
+++ b/tests/test_recon/test_recon.py
@@ -1,49 +1,40 @@
 import cupy as cp
 import numpy as np
 import pytest
-from cupy.testing import assert_allclose
+from numpy.testing import assert_allclose
 
 from httomolib.prep.normalize import normalize_cupy
 from httomolib.recon.rotation import find_center_360, find_center_vo_cupy
 
-in_file = 'tests/test_data/tomo_standard.npz'
-datafile = np.load(in_file)
-# keys: data, flats, darks, angles, angles_total, detector_y, detector_x
-host_data = datafile['data']
-host_flats = datafile['flats']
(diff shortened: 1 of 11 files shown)


tests/test_recon/test_recon.py
```diff
@@ -1,49 +1,40 @@
...
-    cor = find_center_vo_cupy(data)
-    for _ in range(10):
-        assert_allclose(cor, 79.5)
+    cor = find_center_vo_cupy(data).get()
+
+    assert_allclose(cor, 79.5)
+
 
+@cp.testing.gpu
+def test_find_center_vo_cupy(ensure_clean_memory):
```
namannimmo10: let us change the function name here, otherwise, pytest won't collect this.

namannimmo10:
```suggestion
def test_find_center_vo_cupy_ones(ensure_clean_memory):
```

namannimmo10 (approved):
:rocket: thanks

PR #52, opened 2023-02-02
tests: free up GPU memory

namannimmo10:


tests/test_pipeline.py
```diff
@@ -62,3 +67,5 @@ def test_cpu_vs_gpu(host_data, host_flats, host_darks):
 
     #: TEST 4: check if the center of rotation matches for both CPU and GPU
     assert_allclose(tomopy_cor, cor)
+
+    data = flats = darks = data_normalize_cupy = corrected_data = None
```
team-gpu: This is not needed - the fixture is setup now that it runs before the test starts (not after), and those references are gone there anyway. 

If you would like it to cleanup at the end as well, you can modify the `ensure_clean_memory` fixture as:

```python
def ensure_clean_memory():
    cp.get_default_memory_pool().free_all_blocks()
    cp.get_default_pinned_memory_pool().free_all_blocks()
    yield None
    cp.get_default_memory_pool().free_all_blocks()
    cp.get_default_pinned_memory_pool().free_all_blocks()
```

This makes sure that it frees everything before *and* after each test that uses the fixture.

tests/test_prep/test_alignment.py
```diff
@@ -47,3 +47,5 @@ def test_correct_distortion(
 
     assert_allclose(np.mean(corrected_data), mean_value)
     assert np.max(corrected_data) == max_value
+
+    im = None #: free up GPU memory
```
team-gpu: no need - fixture runs after test function is finished, which will have those references removed anyway at that point.

tests/test_prep/test_normalize.py
```diff
@@ -6,7 +6,7 @@
 
 
 @cp.testing.gpu
-def test_normalize_1D_raises(data, flats, darks):
+def test_normalize_1D_raises(data, flats, darks, ensure_clean_memory):
```
team-gpu: also not needed, since `data`, `flats`, `darks` are fixtures that depend on that one already. Check out `dark` in  `conftest.py`.

Any time a GPU data fixture is used, the clean memory is already part of the fixtures implicitly.

team-gpu (changes requested):
Just a note: The `ensure_clean_memory` fixture clears everything allocated with cupy on host (pinned) and device. That's fine if everything is functional and no global data is used (which I believe is good practice anyway). Otherwise there will be dangling references.

tests/test_pipeline.py
namannimmo10: Please read this answer by the `cupy` dev lead: https://stackoverflow.com/a/53537298

xcelerit-team: I understand that. But the fixture executes the memory cleanup *after* the function returns, which means the local variables will already be out of scope and their reference count dropped to zero already before the pool is cleaned. That's why it's not necessary when using a fixture.

namannimmo10: good catch, thanks - I didn't look at this carefully.

PR #53, opened 2023-02-06
Performance tests and more precise assertions for paganin filter

team-gpu:
This adds specifically marked performance tests to the test suite. They are not for functional correctness, but for measuring performance only. They need to be marked with `@pytest.mark.perf`, and will only run with users run `pytest --performance`. The execution time is reported in an assertion, and they will always fail. This allows seeing the execution time nicely in the summary. 

It also adds NVTX annotations to all GPU functions, so they can be highlighted in the profiler. (Note that these annotations do no harm if the program is not executed in a profiler).

Further, it extends one of the `pagainin_filter` tests to not only assert the mean and min, but also some specific values. This is to ensure correctness even if the fftshifts and padding might have gone wrong (the mean might still be matching, while specific value might not in that case).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -26,5 +26,5 @@ Run tests
 * Run GPU tests separately with :code:`$ pytest -v -m gpu`.
 * Run CPU tests separately with :code:`$ pytest -v -m "not gpu"`.
 * Run performance tests (only) with :code:`$ pytest --performance`
-  (note that performance tests always fail - the report the execution time in an assertion
+  (note that performance tests always fail - they report the execution time in an assertion
   to see them in the summary easily)
--- a/README.rst
+++ b/README.rst
@@ -25,3 +25,6 @@ Run tests
 * Run all tests with :code:`$ pytest`. To increase verbosity, use :code:`$ pytest -v`.
 * Run GPU tests separately with :code:`$ pytest -v -m gpu`.
 * Run CPU tests separately with :code:`$ pytest -v -m "not gpu"`.
+* Run performance tests (only) with :code:`$ pytest --performance`
+  (note that performance tests always fail - the report the execution time in an assertion
+  to see them in the summary easily)
--- a/httomolib/prep/phase.py
+++ b/httomolib/prep/phase.py
@@ -147,6 +147,7 @@ def _make_window(height, width, ratio, pattern):
 
 ## %%%%%%%%%%%%%%%%%%%%%%% paganin_filter %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%  ##
 #: CuPy implementation of Paganin filter from Savu
+@nvtx.annotate()
 def paganin_filter(
     data: cp.ndarray,
     ratio: float = 250.0,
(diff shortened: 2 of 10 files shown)


httomolib/prep/phase.py
```diff
@@ -147,6 +147,7 @@ def _make_window(height, width, ratio, pattern):
 
 ## %%%%%%%%%%%%%%%%%%%%%%% paganin_filter %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%  ##
 #: CuPy implementation of Paganin filter from Savu
+@nvtx.annotate()
```
namannimmo10, line 153: you need to import `nvtx` in this file, too

README.rst
```diff
@@ -25,3 +25,6 @@ Run tests
 * Run all tests with :code:`$ pytest`. To increase verbosity, use :code:`$ pytest -v`.
 * Run GPU tests separately with :code:`$ pytest -v -m gpu`.
 * Run CPU tests separately with :code:`$ pytest -v -m "not gpu"`.
+* Run performance tests (only) with :code:`$ pytest --performance`
+  (note that performance tests always fail - the report the execution time in an assertion
+  to see them in the summary easily)
```
namannimmo10:
```suggestion
* Run performance tests (only) with :code:`$ pytest --performance`
  (note that performance tests always fail - they report the execution time in an assertion
  to see them in the summary easily)
```

tests/test_prep/test_phase.py
```diff
@@ -61,6 +63,58 @@ def test_paganin_filter_padmean(data):
...
     assert_allclose(np.mean(filtered_data), -765.3401, rtol=eps)
     assert_allclose(np.min(filtered_data), -793.68787, rtol=eps)
+    # test a few other slices to ensure shifting etc is right
+    assert_allclose(
+        filtered_data[0, 50, 1:5],
+        [-785.60736, -786.20215, -786.7521, -787.25494],
+        rtol=eps,
+    )
+    assert_allclose(filtered_data[0, 50, 40:42], [-776.6436, -775.1906], rtol=eps)
+    assert_allclose(filtered_data[0, 60:63, 90], [-737.75104, -736.6097, -735.49884])
```
namannimmo10: Unfortunately, this test fails on my system, but increasing the epsilon value would make sure it'll pass.
```suggestion
    assert_allclose(filtered_data[0, 60:63, 90],
        [-737.75104, -736.6097, -735.49884], rtol=eps)
```

namannimmo10 (approved):

httomolib/recon/algorithm.py
```diff
@@ -25,13 +25,15 @@
 
 import cupy as cp
 import numpy as np
+import nvtx
```
namannimmo10, line 28: We don't want another dependency, let's use `from cupy.cuda import nvtx`?

namannimmo10: OK, I realized `nvtx` was already in the conda environment because of `cupy`. I made the change in https://github.com/DiamondLightSource/httomolib/pull/54/commits/87cb00ae99ba21cf4d1a7cdcf289e73d9d7d24c5.

PR #55, opened 2023-02-06
Optimised paganin filter for GPU

team-gpu:
Performance (using performance test in the code-base) went from ~1,390ms to ~350ms (~4x faster). Main points:

* filter generation + ifftshift is done in a simple RawKernel
* Pre-processing of data + post-processing done in Elementwise kernel each
* FFT2, filter, IFFT2 is done without for loop, over the whole stack of images

### DIFF
--- a/httomolib/prep/phase.py
+++ b/httomolib/prep/phase.py
@@ -21,8 +21,6 @@
 # ---------------------------------------------------------------------------
 """Modules for phase retrieval and phase-contrast enhancement"""
 
-import math
-
 import cupy as cp
 import cupyx
 import numpy as np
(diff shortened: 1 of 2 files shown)


httomolib/prep/phase.py
```diff
@@ -210,50 +208,124 @@ def paganin_filter(
...
+                cp.float32(wavelength), 
+                cp.float32(distance), 
+                cp.float32(ratio), 
+                filtercomplex))
 
     # Apply padding to all the 2D projections
-    data = cp.pad(data, ((0, 0), (pad_y, pad_y), (pad_x, pad_x)),
-                  mode=pad_method)
+    # Note: this takes considerable time on GPU...
+    data = cp.pad(data, ((0, 0), (pad_y, pad_y), (pad_x, pad_x)), mode=pad_method)
```
namannimmo10, line 261: This seems to be using a lot of GPU memory and hence it crashes on my system!

team-gpu: I have just pushed a change to restrict the memory depending on the GPU type + also reduced memory consumption a bit. Could you give it another try please?

namannimmo10: Nice, now I get:

> `FAILED tests/test_prep/test_phase.py::test_paganin_filter_performance - AssertionError: assert 'performance in ms' == 494.3354851`

httomolib/prep/phase.py
```diff
@@ -210,50 +208,123 @@ def paganin_filter(
...
+        float pxx = (px - centerx) * dpx;
+        float pyy = (py - centery) * dpy;
+        float pd = (pxx*pxx + pyy*pyy) * wavelength * distance * 3.1415926535897932384626433832795f;
+        float filter1 = 1.0f + ratio * pd;
+
+        complex<float> value = 1.0f / complex<float>(filter1, filter1);
+
+        // iffshifting positions
+        int xshift = width1 / 2;
+        if (width1 % 2 != 0) {
```
namannimmo10:
```suggestion
        if (width1 & 1) {
```

httomolib/prep/phase.py
```diff
@@ -210,50 +208,123 @@ def paganin_filter(
...
+            xshift++;
+        }
+        int yshift = height1 / 2;
+        if (height1 % 2 != 0) {
```
namannimmo10:
```suggestion
        if (height1 & 1) {
```

httomolib/prep/phase.py
```diff
@@ -210,50 +208,123 @@ def paganin_filter(
...
+            yshift++;
+        }
+        int outX = (px + xshift) % width1;
+        int outY = (py + yshift) % height1;
+
+        filtercomplex[outY * width1 + outX] = value;
+    }
+    """, "paganin_filter_gen")
+    filtercomplex = cp.empty((height1, width1), dtype=np.complex64)
```
namannimmo10: Just curious, why do you think we need np.complex64 here? Can't `np.complex32` suffice here--we can possibly get a better speedup?

httomolib/prep/phase.py
yousefmoazzam: I'm naive with this, so humour me if this is a silly question: is there any performance reason for using a bitwise operator here? My concern is that the intent of the code as it originally was is clear/readable, but with the `&` it starts to become a bit nebulous (for me at least)?

namannimmo10: Yes, bitwise operations are generally fast

httomolib/prep/phase.py
team-gpu: good - that's how the performance tests work. The assertions are just a hacky way to see the time in the test summary.

httomolib/prep/phase.py
yousefmoazzam: Fair enough, speed is always a worthy reason :smile:

That being said, I'd be interested to see a comparison between the two approach's performance, because there's only two instances of that operation occurring, and it's within a raw CUDA kernel; does the performance of the kernel change significantly between the approach using `%` and `&`? Because, in my opinion, the code readability suffers somewhat in the bitwise operation case, and I'm curious to know how much benefit is gained from that?

team-gpu: I'd be surprised if that makes any difference - they compiler should do these kind of optimisations automatically. It knows it's a modulo-2 operation, so it will very likely replace it by the bit operation anyway.

httomolib/prep/phase.py
```diff
@@ -210,50 +208,123 @@ def paganin_filter(
...
+                            complex<float>* filtercomplex) {
+        int px = threadIdx.x + blockIdx.x * blockDim.x;
+        int py = threadIdx.y + blockIdx.y * blockDim.y;
+        if (px >= width1) return;
+        if (py >= height1) return;
+
+        float dpx = 1.0f / (width1 * resolution);
+        float dpy = 1.0f / (height1 * resolution);
+        float centerx = ceil(width1 / 2.0f) - 1.0f;
+        float centery = ceil(height1 / 2.0f) - 1.0f;
```
namannimmo10:
```suggestion
        float centerx = ceil(width1 * 0.5f) - 1.0f;
        float centery = ceil(height1 * 0.5f) - 1.0f;
```

httomolib/prep/phase.py
team-gpu: `complex64` means both real and imaginary parts are `float32`. I don't think you want half-precision here? https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.csingle

httomolib/prep/phase.py
team-gpu: yes, that can be done. In fact, it's better to do `(width1 + 1) / 2` (integer division, rounding up). I'll push that change.

httomolib/prep/phase.py
namannimmo10: I don't know about your machine, but now I get:

> `FAILED tests/test_prep/test_phase.py::test_paganin_filter_performance - AssertionError: assert 'performance in ms' == 464.45301229999995`

Whereas before the suggestions, it was 494 ms.

namannimmo10: Up to you, if you want this in or not

dkazanc: a noob here. I would keep Yousef's implementation for readability... Regarding the speed, I think we have a bigger fish to fry with the I/O eating up like 90% of all time run :)

httomolib/prep/phase.py
namannimmo10: Ah yes. Anyway, I suggest we use fp16: https://github.com/DiamondLightSource/httomolib/issues/35. So complex32 should fit in right, but again this needs more thought.

namannimmo10: Ah yes. Anyway, I suggest we use fp16: https://github.com/DiamondLightSource/httomolib/issues/35. So complex32 should fit in right, but again this needs more thought.

httomolib/prep/phase.py
team-gpu: just pushed those changes.

httomolib/prep/phase.py
team-gpu: The push I just did avoids this completely - it can in fact be done using `(width1 + 1) / 2` and avoiding the if compleley.

httomolib/prep/phase.py
team-gpu: I see - so far all you have elsewhere is float32 as far as I can see. So that would be a more substantial change - and I'd take that separately.

namannimmo10 (approved):

httomolib/prep/phase.py
dkazanc: I would avoid float16 for now.. it is not supported on all architectures, makes it less generic for others to use.

PR #58, opened 2023-02-07
Median3d cuda kernel in misc.corr

dkazanc:
Adds median3d and outliers removal modules in [misc.corr](https://github.com/DiamondLightSource/httomolib/blob/median/httomolib/misc/corr.py). 
@namannimmo10 Could you add tests please.


httomolib/misc/corr.py
```diff
@@ -23,20 +23,255 @@
...
+    if data.ndim == 3:
+        dz, dy, dx = data.shape
+        if (dz == 0) or (dy == 0) or (dx == 0):
+            raise ValueError("The length of one of dimensions is equal to zero")
+    else:
+        raise ValueError("The input array must be a 3D array")
+
+    median_kernel = r'''
+    template <int radius, int diameter, int midpoint>
+    __global__ void median_float_kernel(const float* in, float* out, float dif, int Z, int M, int N, long num_total)
```
namannimmo10: I guess you can use a type-generic kernel here: https://docs.cupy.dev/en/stable/user_guide/kernel.html#type-generic-kernels. `median_float_kernel` and `median_uint16_kernel` are doing the same computations--they just have different args.

httomolib/misc/corr.py
```diff
@@ -23,20 +23,255 @@
...
+    if input_type not in ["float32", "uint16"]:
+        raise ValueError("The input data should be either float32 or uint16 data type")
+    if dif == 0.0:
+        out = cp.zeros(data.shape, dtype=input_type, order="C")
+    else:
+        out = cp.copy(data, order="C")
+
+    if data.ndim == 3:
+        dz, dy, dx = data.shape
+        if (dz == 0) or (dy == 0) or (dx == 0):
```
namannimmo10:
```suggestion
        if 0 in data.shape:
```

httomolib/misc/corr.py
xcelerit-team: yes, agreed. The kernels are identical, expect for the types. You can use `typename T` in the template parameter, use `T` in the argument list, and instantiate for int16 and float32 as needed.

xcelerit-team (commented):
With the local `[dimension**3]` sized arrays for each GPU thread, you'll very likely get register spilling for any kernel size > 3 or so. That means some of the local variables will actually be stored in memory rather than registers. And other things with the double/tripple loops and memory access patterns will very likely be inefficient too.

I think a completely different design is required for this to make it most efficient. But then again - if it's not a bottleneck and fast enough, it may be ok to leave it.

httomolib/misc/corr.py
dkazanc: Ok, good point! Let me try to adapt it this way

httomolib/misc/corr.py
```diff
@@ -23,20 +23,197 @@
...
+    else:
+        templates = ['median_general_kernel<unsigned short,1,3,13>',
+                'median_general_kernel<unsigned short,2,5,62>',
+                'median_general_kernel<unsigned short,3,7,171>',
+                'median_general_kernel<unsigned short,4,9,364>',
+                'median_general_kernel<unsigned short,5,11,665>',
+                'median_general_kernel<unsigned short,6,13,1098>']
+
+    module = cp.RawModule(code=median_kernel, options=('-std=c++11',),
+                        name_expressions=templates)
```
team-gpu: you only need to compile the template instantiations you actually need. I wouldn't add them all here - just the one you're actually calling. This will make the compilation process faster.

tests/test_misc/test_corr.py
```diff
@@ -15,3 +22,59 @@ def test_inpainting_filter3d(host_data):
...
+@cp.testing.gpu
+def test_median_filter3d_cupy_vs_scipy(host_data, ensure_clean_memory):
+    mat = np.arange(4*5*6).reshape(4, 5, 6)
+    assert_equal(
+        scipy.ndimage.median_filter(np.float32(mat), size=3),
+        median_filter3d_cupy(cp.asarray(mat, dtype=cp.float32), kernel_size=3).get()
+    )
+
+    #: now test with tomo_standard
+    assert_equal(
```
team-gpu, line 42: these should be 2 tests...

yousefmoazzam: I agree with @team-gpu, otherwise if the test fails, it's not clear which data was involved in the failure: the dummy data defined using `np.arange` or `tomo_stadard`?

tests/test_misc/test_corr.py
```diff
@@ -15,3 +22,59 @@ def test_inpainting_filter3d(host_data):
...
+        scipy.ndimage.median_filter(np.float32(host_data), size=3),
+        median_filter3d_cupy(cp.asarray(host_data, dtype=cp.float32), kernel_size=3).get()
+    )
```
namannimmo10:
```suggestion
def test_median_filter3d_cupy_vs_scipy_on_arange(ensure_clean_memory):
    mat = np.arange(4*5*6).reshape(4, 5, 6)
    assert_equal(
        scipy.ndimage.median_filter(np.float32(mat), size=3),
        median_filter3d_cupy(cp.asarray(mat, dtype=cp.float32), kernel_size=3).get()
    )


@cp.testing.gpu
def test_median_filter3d_cupy_vs_scipy(host_data, ensure_clean_memory):
    assert_equal(
        scipy.ndimage.median_filter(np.float32(host_data), size=3),
        median_filter3d_cupy(cp.asarray(host_data, dtype=cp.float32), kernel_size=3).get()
    )
```

httomolib/misc/corr.py
```diff
@@ -23,20 +23,197 @@
...
+__all__ = [    
+    'median_filter3d_cupy',
+    'remove_outlier3d_cupy',
     'inpainting_filter3d',
 ]
 
+def median_filter3d_cupy(data: cp.ndarray,
+                         kernel_size: int = 3,
+                         dif: float = 0.0,
+                         ) -> cp.ndarray:
```
namannimmo10:
```suggestion
def median_filter3d_cupy(
    data: cp.ndarray,
    kernel_size: int = 3,
    dif: float = 0.0
) -> cp.ndarray:
```

httomolib/misc/corr.py
```diff
@@ -23,20 +23,198 @@
...
+    else:
+        raise ValueError("Please select a correct kernel size: 3,5,7,9,11,13")        
+    median3d(grid_dims, block_dims, params)
+    return out
+
+
+def remove_outlier3d_cupy(data: cp.ndarray,
+                         kernel_size: int = 3,
+                         dif: float = 0.1,
+                         ) -> cp.ndarray:
```
namannimmo10:
```suggestion
def remove_outlier3d_cupy(
    data: cp.ndarray,
    kernel_size: int = 3,
    dif: float = 0.1
) -> cp.ndarray:
```

httomolib/misc/corr.py
dkazanc: the users might need larger kernels. Shall we stick to the first 3 templates then?

dkazanc: I need to pass the template I'm using, is that correct?

httomolib/misc/corr.py
```diff
@@ -23,20 +23,200 @@
...
+
+    Raises
+    ------
+    ValueError
+        If the input array is not three dimensional.
+    """                        
+    return median_filter3d_cupy(
+        data = data,
+        kernel_size=kernel_size,
+        dif = dif
```
namannimmo10:
```suggestion
        data=data,
        kernel_size=kernel_size,
        dif=dif
```

tests/test_misc/test_corr.py
```diff
@@ -12,6 +22,104 @@ def test_inpainting_filter3d(host_data):
...
+
+@cp.testing.gpu
+def test_remove_outlier3d_cupy(data):
+    filtered_data = remove_outlier3d_cupy(data, kernel_size=3, dif=1.5).get()
+
+    assert filtered_data.ndim == 3
+    assert_allclose(np.mean(filtered_data), 808.753494, rtol=eps)
+    assert_allclose(np.mean(filtered_data, axis=(1, 2)).sum(), 145575.628906)
+    assert_allclose(np.median(filtered_data), 976.)
+    assert_allclose(np.median(filtered_data, axis=(1, 2)).sum(), 175741.5)
```
namannimmo10:
```suggestion
    filtered_data = remove_outlier3d_cupy(data, kernel_size=9, dif=1.5).get()

    assert filtered_data.ndim == 3
    assert_allclose(np.mean(filtered_data), 809.023452, rtol=eps)
    assert_allclose(np.mean(filtered_data, axis=(1, 2)).sum(), 145624.221436)
    assert_allclose(np.median(filtered_data), 980.)
    assert_allclose(np.median(filtered_data, axis=(1, 2)).sum(), 176400.)
```

namannimmo10:
Merging it for now... we should optimize this in a subsequent PR!

PR #60, opened 2023-02-08
Optimised _filtersinc3D_cupy of reconstruct_tomobar (>100x faster)

team-gpu:
This optimises `reconstruct_tomobar` as far as possible, without touching the astra toolbox call. In particular:

- the FFT2 is done on the whole batch without a for loop
- the filter generation is done in a custom kernel to avoid temporaries and lots of tiny kernel calls
- the axis swapping is done after the filtering, to ensure continuous data for the FFT
- FFT scaling is multiplied into the filter itself, and there's no need to scale after FFT

The performance of `_filtersinc3D_cupy` improves from ~420ms to less than 4ms - greater, see screenshots before/after.

Note: Though we compared the filter visually, there is no real test for this function. So we can't be 100% sure if the result is correct. We suggest that you create a test for it separately, and once it's there, we'll merge it into this branch to confirm.

Before: 
image

After:
image

### DIFF
--- a/httomolib/recon/algorithm.py
+++ b/httomolib/recon/algorithm.py
@@ -24,6 +24,7 @@
 from typing import Optional
 
 import cupy as cp
+import cupyx
 import numpy as np
 import nvtx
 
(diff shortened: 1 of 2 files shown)


httomolib/recon/algorithm.py
```diff
@@ -105,15 +106,16 @@ def reconstruct_tomobar(
                                 device_projector = 'gpu',
                                 GPUdevice_index=gpu_id)    
         # ------------------------------------------------------- # 
-        data = _filtersinc3D_cupy(cp.swapaxes(data, 0, 1)) # filter the data on the GPU and keep the result there        
-        
+        data = _filtersinc3D_cupy(data) # filter the data on the GPU and keep the result there
+        data = cp.swapaxes(data, 0, 1)
```
namannimmo10: You are changing the order of the function call, aren't you? Looks like we don't have a test for `FBP3D_device` yet.

team-gpu: Yes, that test is missing (see the PR description). Since `_filtersync3D_cupy` is private and only called from here, I saw that it works faster with the original axes. Only the astra toolbox needs them swapped, so I do the swapping afterwards.

namannimmo10:
Sorry, you need to resolve the conflicts--rebase with `master`.

PR #61, opened 2023-02-08
Add tests for reconstruction

namannimmo10:


tests/test_recon/test_algorithm.py
```diff
@@ -1,30 +1,80 @@
...
     assert_allclose(np.mean(recon_data_tomopy), 0.008697214, rtol=1e-07, atol=1e-8)
     assert_allclose(np.std(recon_data_tomopy), 0.009089365, rtol=1e-07, atol=1e-8)
 
     #: check that the reconstructed data is of type float32
-    assert recon_data.dtype == np.float32
     assert recon_data_tomopy.dtype == np.float32
+
+
+@cp.testing.gpu
+def test__filtersinc3D_cupy(data, ensure_clean_memory):
```
team-gpu: Do you consider `_filtersinc3D_cupy` a public function, part of the interface? The leading underscore suggests that it's private, and might change or be renamed any time. Tests shouldn't test private functions...

tests/test_recon/test_algorithm.py
```diff
@@ -1,30 +1,80 @@
...
+    host_darks,
+    ensure_clean_memory
+):
+    data = normalize(host_data, host_flats, host_darks, cutoff=15.0)
+    angles = np.linspace(0. * np.pi / 180., 180. * np.pi / 180., data.shape[0])
+
+    recon_data_tomopy = reconstruct_tomopy(data, angles, 79.5, algorithm="FBP_CUDA")
+
     assert_allclose(np.mean(recon_data_tomopy), 0.008697214, rtol=1e-07, atol=1e-8)
     assert_allclose(np.std(recon_data_tomopy), 0.009089365, rtol=1e-07, atol=1e-8)
```
team-gpu, line 95: in order to ensure the shifting etc is done correctly, I'd also add 1 or 2 elements somewhere off center and verify those - or the mean along an off-centre axis or something like that. Otherwise if the whole array came back shifted (e.g. mistake in an fftshift), your test would still pass.

team-gpu (changes requested):

tests/test_recon/test_algorithm.py
```diff
@@ -1,30 +1,80 @@
...
+def test_reconstruct_tomopy_fbp_cuda(
+    host_data,
+    host_flats,
+    host_darks,
+    ensure_clean_memory
+):
+    data = normalize(host_data, host_flats, host_darks, cutoff=15.0)
+    angles = np.linspace(0. * np.pi / 180., 180. * np.pi / 180., data.shape[0])
+
+    recon_data_tomopy = reconstruct_tomopy(data, angles, 79.5, algorithm="FBP_CUDA")
```
team-gpu, line 90: I believe it's `FBP3D_device` - from the code it looks like it supports `FBP3D_host` and `FBP3D_device`

team-gpu: And if you set it to `FBP3D_device`, you should get back a cupy array, and the whole test should be marked `@cp.testing.gpu`

tests/test_recon/test_algorithm.py
namannimmo10: How do you suggest we test this function then? :)

team-gpu: it's tested with the `FBP3D_device` test

tests/test_recon/test_algorithm.py
team-gpu: sorry - I got confused with the naming. I mean we need a test which calls `reconstruct_tomobar` and `algorithm='FBP3D_device'`. That's missing. (this test is calling `reconstruct_tomopy`, so ignore my comments above...)

tests/test_recon/test_algorithm.py
```diff
@@ -1,30 +1,80 @@
...
+    recon_data = reconstruct_tomobar(
+        normalize_cupy(data, flats, darks, cutoff=10, minus_log=True),
+        angles,
+        79.5,
+        algorithm="FBP3D_device"
+    )
+
+    assert_allclose(np.mean(recon_data), 0.000798, rtol=1e-07, atol=1e-6)
+    assert_allclose(np.std(recon_data), 0.006293, rtol=1e-07, atol=1e-6)
+    assert recon_data.dtype == np.float32
```
namannimmo10: > I mean we need a test which calls `reconstruct_tomobar` and `algorithm='FBP3D_device'`. That's missing.

It is right here, innit?

team-gpu: I must be dreaming... it is indeed. Sorry for all that.

namannimmo10:
@dkazanc -- `FBP_CUDA` in

https://github.com/DiamondLightSource/httomolib/blob/ba746a621a783b8c5a575738cf8788cf7c23201d/httomolib/recon/algorithm.py#L161

is not a valid algorithm in https://github.com/tomopy/tomopy/blob/c236a2969074f5fc70189fb5545f0a165924f916/source/tomopy/recon/algorithm.py#L129... any reason why you added that?

tests/test_recon/test_algorithm.py
namannimmo10: No worries!

tests/test_recon/test_algorithm.py
namannimmo10: I added this to be completely sure that we get the right output after optimisation, but fine, I can remove this test. Thanks.

dkazanc:
_FBP_CUDA_ method it is ASTRA related thing in TomoPy https://tomopy.readthedocs.io/en/latest/ipynb/astra.html

yousefmoazzam:
@namannimmo10 I think it's a case where the docstring of a function (in this case, TomoPy's `recon()` function) is somewhat incomplete, otherwise things become too verbose; note that the type for the `algorithm` parameter is `{str, function}`, so it at least _hints_ that it's maybe not as simple as it taking only a `str`.

The `algorithm` parameter of `tomopy.recon()` is a `str` when specifying non-Astra algorithms I think.

However, I think one can also pass the `tomopy.astra` module (ie, `algorithm=tomopy.astra`) as is done in the `reconstruct_tomopy()` function in httomo. This is to then specify different algorithms that run on GPU that have been implemented in Astra, by using the `options` keyword arg, as shown in the link that Daniil posted.

The `FPB_CUDA` that you mentioned is then specifying the algorithm that will end up in the `options` dict.

namannimmo10:
I see, thanks Yousef and Daniil... I am not touching `reconstruct_tomopy` in this PR although it can be improved.

namannimmo10:
I think this should be fine to complete #60! Let's add more tests in subsequent PRs.

PR #66, opened 2023-02-20
Optimised find_center_vo_cupy

team-gpu:
This optimises the function `find_center_vo_cupy` for the GPU, as far as it goes without changing the algorithm / logic involved. 

### Performance Improvement: 255ms -> 194ms (1.31x)

(This was measured on V100 using the performance test that is part of this PR.)

### Remaining Bottlenecks

The key bottleneck however are the spline shifts for sub-pixel shifting - they dominate the overall time by far (ignoring the FFT), especially after these optimisations. We get the following picture for the kernel percentages:

| Function | Percent GPU before | Percent GPU after |
|---|---:|---:|
| FFT | 44.3% | 46.4% |
| Spline Shift | 29.8% | 33.6% |
| all other kernels combined | 20.4% | 19.2% |
| Memory operations | 5.5% | 0.8% |

(Note that the FFT dimensions are not powers of 2 - this causes the long FFT time. Also note that a big part of the optimisation was reducing the number of kernel calls and hereby reducing gaps where the GPU is idle.)

### Details of Changes

- small operations such as ceil, div, abs, of scalars are computed on the CPU instead of launching kernels
- other small operations where joined on the GPU
- axis extensions and copies of data are avoided
- a single RawKernel does:
   - integer shifts of the input data for all lines where the shift is integer
   - inserts the `sino3` array into the part which was shifted out
 - then the scipy ndimage shift function (using order 3 splines with prefiltering) is only called for the rows fractional shifts, in a loop
 - the FFT, masking, and mean calculation is done once for all possible shifts (outside of the loop)
 - the downsample and mask generation kernels have been optimised further

### DIFF
--- a/httomolib/recon/rotation.py
+++ b/httomolib/recon/rotation.py
@@ -21,6 +21,7 @@
 # ---------------------------------------------------------------------------
 """Modules for finding the axis of rotation"""
 
+import math
 from typing import Optional
 
 import cupy as cp
@@ -98,13 +99,15 @@ def _find_center_vo_gpu(sino, ind, smin, smax, srad, step, ratio, drop):
     else:
         _sino = sino[:, ind, :]
 
-    _sino_cs = gaussian_filter(_sino, (3, 1), mode="reflect")
-    _sino_fs = gaussian_filter(_sino, (2, 2), mode="reflect")
-
+    with nvtx.annotate("gaussian_filter_1", color="green"):
+        _sino_cs = gaussian_filter(_sino, (3, 1), mode="reflect")
+    with nvtx.annotate("gaussian_filter_2", color="green"):
+        _sino_fs = gaussian_filter(_sino, (2, 2), mode="reflect")
+    
     if _sino.shape[0] * _sino.shape[1] > 4e6:
         # data is large, so downsample it before performing search for
         # centre of rotation
-        _sino_coarse = _downsample(cp.expand_dims(_sino_cs, 1), 2, 2)[:, 0, :]
+        _sino_coarse = _downsample(_sino_cs, 2, 1)
         init_cen = _search_coarse(_sino_coarse, smin / 4.0, smax / 4.0, ratio, drop)
         fine_cen = _search_fine(_sino_fs, srad, step, init_cen * 4.0, ratio, drop)
     else:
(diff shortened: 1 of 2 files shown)


httomolib/recon/rotation.py
```diff
@@ -242,45 +252,119 @@ def _create_mask(nrow, ncol, radius, drop):
...
-    # better performance
-    if cp.abs(shift_col - cp.floor(shift_col)) == 0.0:
-        shift_col = int(shift_col)
-        sino_shift = cp.roll(sino2, shift_col, axis=1)
-        if shift_col >= 0:
-            sino_shift[:, :shift_col] = sino3[:, :shift_col]
-        else:
-            sino_shift[:, shift_col:] = sino3[:, shift_col:]
-        mat = cp.vstack((sino1, sino_shift))
+def round_up(x: float) -> int:
```
namannimmo10, line 262: you can use python's `round` function here.

team-gpu: Unfortunately not - somehow the original rounding is not standard. See here for comparison:

```
>>> [round_up(x) for x in [0.0, 1.2, 2.5, -1.2, -1.6]]
[0, 2, 3, -2, -2]
>>> [round(x) for x in [0.0, 1.2, 2.5, -1.2, -1.6]]   
[0, 1, 2, -1, -2]
```

It rounds away from zero. See original code, lines 269 and 272

namannimmo10: I see -  then it is fine to have this.

httomolib/recon/rotation.py
```diff
@@ -242,45 +252,119 @@ def _create_mask(nrow, ncol, radius, drop):
...
+            if shift_int >= 0:
+                mat[i, na1:, shift_int:] = shifted[:, shift_int:]
+            else:
+                mat[i, na1:, :shift_int] = shifted[:, :shift_int]
+
+        
+    # stack and transform
+    mat[:,:na1,:] = sino1
+    cp.abs(cp.fft.fft2(mat, axes=(1,2)), out=mat)
+    mat *= mask
```
namannimmo10, line 358: Do you think taking the `abs` and doing the multiplication here in a separate `ElementwiseKernel` might speed things up a little?

team-gpu: You are right - an element-wise kernel would combine these two operations and will be a bit faster. But in the big picture, we're talking about microseconds here, and it doesn't save memory either. The profiler didn't show these calls as significant at all. I don't think that's worth doing - the benefit will not be noticeable. But happy to put it in if you prefer.

namannimmo10:
Should we merge this then?

PR #67, opened 2023-02-20
Further optimisation of normalise kernel

team-gpu:
This adds some further optimisations on the normalise kernel. 

### Details:

- avoid making data a float, as it's not necessary (allow different types for inputs of the elementwise kernel)
- avoid temporaries + copies by preallocating and writing results of mean directly into the memory given
- build the kernel string dynamically based on the nonegative, log, and nan removal flags (avoiding unnecessary conditionals in the kernel and better optimisations)
- avoid writing back to global memory in every step - use a local variable `v` instead

### DIFF
--- a/httomolib/prep/normalize.py
+++ b/httomolib/prep/normalize.py
@@ -31,6 +31,7 @@ __all__ = [
 ]
 
 
+@nvtx.annotate()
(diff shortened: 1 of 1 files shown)


namannimmo10:
Great, many thanks.

PR #68, opened 2023-02-21
Make `remove_stripe_titarenko` CPU-GPU agnostic

namannimmo10:
Made the name similar to TomoPy.

### DIFF
--- a/httomolib/prep/stripe.py
+++ b/httomolib/prep/stripe.py
@@ -20,6 +20,8 @@
 # version ='0.1'
 # ---------------------------------------------------------------------------
 """Modules for stripes removal"""
+from typing import Union
+
 import cupy as cp
 import numpy as np
 import nvtx
@@ -111,9 +113,9 @@ def _rs_sort(sinogram, size, dim):
 
 @nvtx.annotate()
 def remove_stripe_ti(
-    data: ndarray,
+    data: Union[cp.ndarray, np.ndarray],
     beta: float = 0.1,
-) -> ndarray:
+) -> Union[cp.ndarray, np.ndarray]:
     """
     Removes stripes with the method of V. Titarenko (TomoCuPy implementation)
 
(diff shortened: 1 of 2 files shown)


httomolib/prep/stripe.py
```diff
@@ -109,12 +109,11 @@ def _rs_sort(sinogram, size, dim):
...
 
 
-## %%%%%%%%%%%%%%%%%%% remove_stripes_titarenko_cupy %%%%%%%%%%%%%%%%%%%%%%%  ##
 @nvtx.annotate()
-def remove_stripes_titarenko_cupy(
+def remove_stripe_ti(
     data: ndarray,
     beta: float = 0.1,
-) -> np.ndarray:
+) -> ndarray:
```
team-gpu: you can use `typings.Union` to give multiple possible types. like this:

```python
from typings import Union

def remove_stripe_ti(data: Union[cp.ndarray, np.ndarray], ...)
```

tests/test_prep/test_stripe.py
```diff
@@ -6,26 +6,37 @@
...
     remove_stripe_based_sorting_cupy,
-    remove_stripes_titarenko_cupy,
+    remove_stripe_ti,
 )
 from numpy.testing import assert_allclose
 
 
 @cp.testing.gpu
-def test_stripe_removal_titarenko_cupy(data, flats, darks):
+def test_remove_stripe_ti_on_data(data, flats, darks):
```
team-gpu, line 15: I think now that it works with both numpy + cupy, why not add another test which generates random data and compares numpy to cupy output? With this: https://docs.cupy.dev/en/stable/reference/generated/cupy.testing.numpy_cupy_allclose.html

team-gpu (changes requested):

httomolib/prep/stripe.py
namannimmo10: Fair point. This can be helpful for readability, although python doesn't do strict type-checking.

team-gpu: IDEs typically underline if the types don't match, and it helps with syntax completion. We generally find strict type specifiers very helpful.

namannimmo10: Oh yes, strict type-checking is the best, helps in catching bugs easily.

PR #71, opened 2023-02-27
Make `remove_stripe_based_sorting` CPU-GPU agnostic

namannimmo10:


tests/test_prep/test_stripe.py
```diff
@@ -54,25 +54,38 @@ def test_remove_stripe_ti_numpy_vs_cupy_on_random_data():
...
+def test_stripe_removal_sorting_numpy_vs_cupy_on_random_data():
+    host_data = np.random.random_sample(size=(181, 5, 256)).astype(np.float32) * 2.0
+    corrected_host_data = remove_stripe_based_sorting(np.copy(host_data))
+    corrected_data = remove_stripe_based_sorting(cp.copy(
+        cp.asarray(host_data))).get()
+
+    assert_allclose(
+        np.sum(corrected_data), np.sum(corrected_host_data))
+    assert_allclose(
+        np.median(corrected_data), np.median(corrected_host_data), rtol=1e-6)
```
team-gpu: I think here you can actually compare the full array, not just the median and sum. Both CPU and GPU should return matching arrays.

PR #72, opened 2023-02-27
Unified numpy/cupy test infrastructure

team-gpu:
This PR introduces a few test fixtures that allow to write tests both for GPU and CPU in a unified manner, and adds them to the `find_center_360` method tests. Note that the actual GPU implementation of that method is not part of the PR, so some tests raise a `NotImplementedError` for now. 

It also adds a pattern for a unity test, which compares the numpy with the cupy results and is parameterised on all possible flags.

This is how it works:

### Test that should be skipped if no GPU is available:

```python

def test_some_gpu_only_function(have_gpu):
    ...
```

### Test that should be run twice - once with and once without GPU:

```python

def test_gpu_cpu_agnostic_function(gpu):
    xp = cp if gpu else np
    data = xp.ones(...)
    
    output = function_under_test(data)
```

### DIFF
--- a/tests/test_recon/test_recon.py
+++ b/tests/test_recon/test_recon.py
@@ -1,10 +1,10 @@
 import time
 import cupy as cp
-import nvtx
+from cupy.cuda import nvtx
 import numpy as np
 import pytest
 from httomolib.prep.normalize import normalize_cupy
-from httomolib.recon.rotation import find_center_360, find_center_vo_cupy, find_center_360_cupy
+from httomolib.recon.rotation import find_center_360, find_center_vo_cupy
 from numpy.testing import assert_allclose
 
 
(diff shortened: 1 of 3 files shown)


tests/test_recon/test_recon.py
```diff
@@ -31,54 +31,112 @@ def test_find_center_vo_cupy_ones(ensure_clean_memory):
...
+
+    return xp.asarray([cor, overlap, side, overlap_pos])
+    
 
 
-@cp.testing.gpu
 @pytest.mark.perf
-def test_find_center_vo_cupy_performance():
+def test_find_center_360_performance(gpu, ensure_clean_memory):
+    xp = cp if gpu else np
```
namannimmo10, line 107: Ahh so you're comparing numpy vs cupy performance here... nice!

team-gpu: yes - then you get a response like this:

```
FAILED tests/test_recon/test_recon.py::test_find_center_360_performance[cupy] - AssertionError: assert 'performance in ms' == 7.2750919
FAILED tests/test_recon/test_recon.py::test_find_center_360_performance[numpy] - AssertionError: assert 'performance in ms' == 1563.3967386
```

team-gpu:
We'll close this one, as we wanted to include it in the other. It makes more sense, as this one ends up with failing tests in master.

PR #73, opened 2023-02-28
GPU ported and optimised find_center_360: >420x speedup

team-gpu:
## Overview

This PR ports the `find_center_360` function to the GPU using cupy, and optimises the function on both CPU and GPU. It also adds appropriate tests and compares it to the CPU reference in all combinations of flags.

## Performance (426x faster)

Based on the performance test in this PR, with normalisation, filtering, and overlap enabled. Measured on Diamond cluster, V100

* Starting performance CPU: **~3,070ms**
* Straight port to Cupy, using `get_array_module`: **~7,150ms**
* Final performance CPU: **~1,580ms**
* Final performance GPU: **~7.2ms** (426x faster than initial CPU, 993x faster than straight GPU port)

## Details

* The key item was to put the calculation of all the correlation metrics (with all possible position shifts) into a single kernel
* The Pearson correlation coefficient can in fact be calculated easily by hand, for every position in a different thread block
* Also removed a lot of temporary arrays and small operations, which also benefited the CPU implementation
* Apart from that, unified tests have been added and the CUDA code has been moved into its own folder, which makes development much easier and code better organised

### DIFF
--- a/httomolib/recon/rotation.py
+++ b/httomolib/recon/rotation.py
@@ -22,15 +22,16 @@
 """Modules for finding the axis of rotation"""
 
 import math
-from typing import Optional
+from typing import Optional, Tuple, Union
 
 import cupy as cp
 import numpy as np
 import nvtx
-import scipy.ndimage as ndi
+import cupyx.scipy.ndimage as cpndi
 from cupy import ndarray
 from cupyx.scipy.ndimage import gaussian_filter, shift
-from scipy import stats
+
+from httomolib.cuda_kernels import load_cuda_module
 
 __all__ = [
     'find_center_vo_cupy',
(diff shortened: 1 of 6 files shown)


httomolib/recon/rotation.py
```diff
@@ -682,16 +780,17 @@ def _calculate_curvature(list_metric):
     min_pos : float
         Position of the minimum value with sub-pixel accuracy.
     """
+    xp = cp.get_array_module(list_metric)
```
namannimmo10: why do you need xp here?

team-gpu:
Let me know if you have further questions / comments on this PR.

team-gpu:
I've refactored this PR as discussed - now the production code is GPU only, and the CPU code has been moved into the tests for reference purposes only. Note that the method name has been changed to add the suffix `_cupy` - not sure if that's desired?

tests/test_recon/test_recon.py
```diff
@@ -31,19 +32,44 @@ def test_find_center_vo_cupy_ones(ensure_clean_memory):
     mat = None #: free up GPU memory
 
 
+@cp.testing.gpu
+@pytest.mark.perf
+def test_find_center_vo_cupy_performance():
```
namannimmo10: We already have this in `master` -- rebase with it, please.

dkazanc:
> I've refactored this PR as discussed - now the production code is GPU only, and the CPU code has been moved into the tests for reference purposes only. Note that the method name has been changed to add the suffix `_cupy` - not sure if that's desired?

Thanks! I think the suffix `cupy` is now redundant and can be dropped for all methods.

tests/test_recon/test_recon.py
team-gpu: I have rebased it now - the reason this shows is that I moved it higher up - from the bottom of the file to this place. That's to keep the `vo_cupy` and the `360`  tests together.

team-gpu:
> > I've refactored this PR as discussed - now the production code is GPU only, and the CPU code has been moved into the tests for reference purposes only. Note that the method name has been changed to add the suffix `_cupy` - not sure if that's desired?
> 
> Thanks! I think the suffix `cupy` is now redundant and can be dropped for all methods.

I removed it for this PR. We'll need to do a global refactoring to make all methods and tests consistent with this strategy after.

PR #74, opened 2023-03-02
Further optimisations on median_filter3d

team-gpu:
This reduces the GPU time of the `median_filter3d_cupy` function from 5.8ms to 3.2ms (for the performance test with kernel size 3 (radius=1). 

The block sizes and register usage has been optimised to achieve a higher occupancy, and some small cupy operations are now performed on the host (calls to `cp.ceil` were running on GPU).

Note that for larger kernel sizes, the local array explodes in size and the registers per thread becomes very large. This limits GPU occupancy and slows down the kernel significantly. However, without dramatically changing the kernel structure and increasing memory a lot, this cannot be avoided. For example, moving from kernel size 3 to kernel size 5 triples the runtime To 10ms, and increasing to 7 brings the runtime to 2000ms. Anything more is prohibitive.

Also, we cannot reduce the memory requirement by working in-place, as this would create a race condition (outputs written by some threads affects inputs of other threads).

### DIFF
--- a/httomolib/misc/corr.py
+++ b/httomolib/misc/corr.py
@@ -64,100 +64,97 @@ def median_filter3d_cupy(
     if input_type not in ["float32", "uint16"]:
         raise ValueError("The input data should be either float32 or uint16 data type")
 
-    if dif == 0.0:
-        out = cp.zeros(data.shape, dtype=input_type, order="C")
-    else:
-        out = cp.copy(data, order="C")
-
+    
     if data.ndim == 3:
         if 0 in data.shape:
             raise ValueError("The length of one of dimensions is equal to zero")
     else:
         raise ValueError("The input array must be a 3D array")
 
+    out = cp.empty(data.shape, dtype=input_type, order="C")
+    
     median_kernel = r'''
-        template <typename Type, int radius, int diameter, int midpoint>
+        template <typename Type, int diameter>
         __global__ void median_general_kernel(
-            const Type* in, Type* out, float dif, int Z, int M, int N, long num_total)
+            const Type* in, Type* out, float dif, int Z, int M, int N)
         {
-            Type ValVec[diameter*diameter*diameter];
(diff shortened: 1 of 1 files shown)


httomolib/misc/corr.py
```diff
@@ -64,100 +64,97 @@ def median_filter3d_cupy(
...
+                            k1 = k;
+                        ValVec[counter] = in[i1 + N*j1 + N*M*k1];
+                        counter++;
                     }
                 }
-                /* do bubble sort here */
-                for (x = 0; x < diameter*diameter*diameter - 1; x++)
+            }
+
+            /* do bubble sort here */
```
namannimmo10, line 115: should we try a better sorting algorithm here?

team-gpu: I had tried shell-sort, and it was actually worse. Recursive algorithms don't work well on GPUs either - but it can very likely be improved, yes. Just wondering about diminishing returns here - if kernel size 3 is common, it's probably not worth it. If you try to use 7 or more, then we should really find a different way to implement this kernel as a whole.

namannimmo10: I see. Did you try `kernel_size >= 9`? Wondering if it works properly or not.

namannimmo10 (approved):

httomolib/misc/corr.py
namannimmo10: I saw this median filter cuda implementation which is doing a similar sorting:
https://github.com/jonaylor89/Median-Filter-CUDA/blob/88ab04b83b89ec6c5a1ec5ed73af474f67100de6/main.cu#L93-L102

dkazanc: that's the bubble sorting, yes. I considered the qsort but as @team-gpu said moved away because of the recursive nature.  I don't think that people would use very large kernels > 7, so in general we should be OK.

PR #78, opened 2023-03-07
Add conda recipe

namannimmo10:


dkazanc:
previously `cuda_kernels` files were not copied during the recipe build. It has been fixed with added MANIFEST.in file. However, the build still fails on `import httomolib` and it is extremely puzzling...

PR #82, opened 2023-03-15
Fix indentation inside method, remove duplicated tests

namannimmo10:


httomolib/recon/algorithm.py
```diff
@@ -67,35 +67,42 @@ def reconstruct_tomobar(
...
+    # initiate a 3D ASTRA class object
+    Atools = AstraTools3D(
+        DetectorsDimH=data.shape[2],  # DetectorsDimH # detector dimension (horizontal)
+        DetectorsDimV=data.shape[
+            1
+        ],  # DetectorsDimV # detector dimension (vertical) for 3D case only
+        AnglesVec=-angles,  # the vector of angles in radians
+        CenterRotOffset=data.shape[2] / 2
+        - center
+        - 0.5,  # The center of rotation combined with the shift offsets
```
namannimmo10:
```suggestion
        # DetectorsDimV: detector dimension (vertical) for 3D case only
        DetectorsDimV=data.shape[1],
        AnglesVec=-angles,  # the vector of angles in radians
        # The center of rotation combined with the shift offsets
        CenterRotOffset=data.shape[2] / 2 - center - 0.5,
```

httomolib/recon/algorithm.py
```diff
@@ -67,35 +67,40 @@ def reconstruct_tomobar(
...
+    # the Astra toolbox requires C-contiguous arrays, and swapaxes seems to return a sliced view which
+    # is neither C nor F contiguous.
+    # So we have to make it C-contiguous first
+    data = cp.ascontiguousarray(cp.swapaxes(data, 0, 1))
+    reconstruction = Atools.backprojCuPy(
+        data
+    )  # backproject the filtered data while keeping data on the GPU
+    cp._default_memory_pool.free_all_blocks()
+    # ------------------------------------------------------- #
     return reconstruction
```
namannimmo10:
```suggestion
    # filter the data on the GPU and keep the result there
    data = _filtersinc3D_cupy(data)

    # the Astra toolbox requires C-contiguous arrays, and swap axes seem to return a sliced view which
    # is neither C nor F contiguous.
    # So we have to make it C-contiguous first
    data = cp.ascontiguousarray(cp.swapaxes(data, 0, 1))

    # backproject the filtered data while keeping data on the GPU
    reconstruction = Atools.backprojCuPy(data)
    cp._default_memory_pool.free_all_blocks()

    return reconstruction
```

PR #84, opened 2023-03-23
Adds sino_360_to_180 method 

team-gpu:
This adds a cupy version of the `sino_360_to_180` method [from tomopy](https://github.com/tomopy/tomopy/blob/master/source/tomopy/misc/morph.py#L314). It tests it against tomopy in all parameter configurations, and also performs tests for invalid inputs. 

The performance test shows in GPU that the function is likely fast enough on V100 without further optimisations. It could possibly be accelerated further with a `RawKernel`, which can combine all the 3 lines of the algorithm and avoid the space for the `linspace` - at the expense of readability.

The performance tests show the following on the Hamilton cluster with V100:

- `rotation=left`: 24.3ms  (vs 2997ms with tomopy) 
- `rotation=right`: 21.1ms (vs 2826ms with tomopy)

So it is around **130x faster than tomopy**

### DIFF
--- a/httomolib/misc/morph.py
+++ b/httomolib/misc/morph.py
@@ -54,6 +54,9 @@ def sino_360_to_180(
     cp.ndarray
         Output 3D data.
     """
+    if data.ndim != 3:
+        raise ValueError("only 3D data is supported")
+    
     dx, dy, dz = data.shape
 
     overlap = int(np.round(overlap))
--- a/httomolib/misc/morph.py
+++ b/httomolib/misc/morph.py
@@ -62,6 +62,8 @@ def sino_360_to_180(
     overlap = int(np.round(overlap))
     if overlap >= dz:
         raise ValueError("overlap must be less than data.shape[2]")
+    if overlap < 0:
+        raise ValueError("only positive overlaps are allowed.")
 
     n = dx // 2
 
--- a/httomolib/misc/morph.py
+++ b/httomolib/misc/morph.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# ---------------------------------------------------------------------------
+# Copyright 2023 Diamond Light Source Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


httomolib/misc/morph.py
```diff
@@ -0,0 +1,86 @@
...
+        Overlapping number of pixels.
+    rotation : string, optional
+        'left' if rotation center is close to the left of the
+        field-of-view, 'right' otherwise.
+    Returns
+    -------
+    cp.ndarray
+        Output 3D data.
+    """
+    dx, dy, dz = data.shape
```
namannimmo10, line 60: check if `data` is 3d?

httomolib/misc/morph.py
```diff
@@ -0,0 +1,86 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ---------------------------------------------------------------------------
+# Created By  : Tomography Team at DLS <scientificsoftware@diamond.ac.uk>
+# Created Date: 23 March 2023
+# version ='0.1'
+# ---------------------------------------------------------------------------
+"""Module for data type morphing functions"""
```
namannimmo10, line 22: please import this module under `__init__.py`

tests/test_misc/test_morph.py
```diff
@@ -0,0 +1,78 @@
...
+        (0, "invalid"),
+        (0, ""),
+    ],
+)
+@cp.testing.gpu
+def test_sino_360_to_180_invalid(ensure_clean_memory, overlap, rotation):
+    data = cp.ones((10, 10, 10), dtype=cp.float32)
+
+    with pytest.raises(ValueError):
+        sino_360_to_180(data, overlap, rotation)
```
namannimmo10, line 47: It is not clear to me why this would raise a `ValueError`? can you add a comment?

team-gpu: This is a test passing invalid combinations of overlap and rotation parameters, such as negative overlaps or rotations that are not 'left' or 'right'. Each one of them should raise an exception. I think since the test has `invalid` in its name, that should be clear enough, but I can add a comment if you prefer?

httomolib/misc/morph.py
team-gpu: done

PR #89, opened 2023-04-05
GPU Memory workaround work

dkazanc:
from @team-gpu :

This adds memory estimation functions and decorators for the methods, including tests. This is documented in the doc/decorators.rst file as well as docstrings.

What is still missing:
- separate tests for decorators / memory estimation from the functional tests (example done in test_alignment.py)
- some methods are still missing these memory estimator tests. Adding them will reveal mistakes in the functions themselves
- possibly adapting the estimator functions to also return the output datatype, as that might change depending on inputs and parameters
- documentation cleanup


httomolib/recon/algorithm.py
```diff
@@ -65,92 +90,196 @@ def reconstruct_tomobar(
...
+    # a guess for astra toolbox memory usage for projection/backprojection
+    astra_size = 0.5*(x_rec+data_out)
+   
+    total_mem = int(data_out + x_rec + d + r + Ad + s + astra_size)
+    slices_max = available_memory // total_mem
+    return (slices_max, float32())
+## %%%%%%%%%%%%%%%%%%%%%%% CGLS reconstruction %%%%%%%%%%%%%%%%%%%%%%%%%%%%  ##
+@method_sino(_calc_max_slices_CGLS)
+@nvtx.annotate()
+def CGLS_rec(
```
namannimmo10, line 219: add this to `__all__`?

tests/test_recon/rotation_cpu_reference.py
```diff
@@ -36,7 +37,7 @@ def find_center_360_numpy(
     else:
         cor = ncol - overlap / 2.0 - 1.0
 
-    return np.float32(cor), np.float32(overlap), side, np.float32(overlap_position)
+    return float(cor), float(overlap), side, float(overlap_position)
```
namannimmo10, line 40: Note that `float` in Python is equivalent to `np.float64`, is that what you want?

tests/test_recon/test_algorithm.py
```diff
@@ -28,31 +32,223 @@ def test_reconstruct_tomobar_device_1(data, flats, darks, ensure_clean_memory):
...
+    objrecon_size = data.shape[2]    
+    cp.get_default_memory_pool().free_all_blocks()
+    cache = cp.fft.config.get_plan_cache()
+    cache.clear()
 
+    objrecon_size = data.shape[2]
+    hook = MaxMemoryHook(data.size * data.itemsize)
+    with hook:
+        recon_data = CGLS_rec(
+            data,
```
namannimmo10, line 233: here we are passing `data` without normalizing it, we might get the same issue with `NaN`s before...

dkazanc: for `MaxMemoryHook` we are not interested in the result, just the memory usage, so it should be ok here. We normalise for non-hook methods.

tests/test_recon/rotation_cpu_reference.py
dkazanc: it is a scalar, but probably keeping the 32 bit consistency makes sense

PR #93, opened 2023-04-24
working conda recipe

dkazanc:
OK, many thanks @yousefmoazzam to sort it out! 
The recipe is working now (expectantly failing on recon tests) and I've added info how to run it in the readme file. 

I think this can be merged as this is an independent PR.


dkazanc:
@namannimmo10 potentially this can be also merged into `memory` branch where tests pass now after `recon` [PR](https://github.com/DiamondLightSource/httomolib/pull/94). As `memory` branch needs constant testing as we merge into it, the idea may be to run Jenkins tests on it nightly (when Jenkins is ready). Just an idea though...

PR #100, opened 2023-05-30
[WIP] Changes necessary for httomo to run with only CPU dependencies

yousefmoazzam:
Motivated by https://github.com/DiamondLightSource/httomo/pull/136, and is largely experimental still.

I'm open to how this could be done (in particular, how the dependencies of httomolib and httomo are defined between conda and pip), this is only one way to enabling the httomo PR to work, so suggestions are most welcome :)


yousefmoazzam:
After some discussion we decided to go a different route to enable httomo to run with only CPU dependencies, which will be cleaner in the long run. Check out the repo of CPU-only methods https://github.com/DiamondLightSource/httomolib that will be part of the new solution. Closing this PR and keeping it as a reference.

PR #103, opened 2023-06-28
Add GPU implementation for paganin filter

namannimmo10:


httomolibgpu/prep/phase.py
```diff
@@ -592,5 +596,133 @@ def _reciprocal_coord(pixel_size: float, num_grid: int) -> cp.ndarray:
...
+
+    # Perform padding to the power of 2 as FFT is O(n*log(n)) complexity
+    # NOTE: Need to convert to float32 as FFT produces complex128 array from uint16
+    # TODO: adding other options of padding?
+    padded_tomo, pad_tup = _pad_projections_to_second_power(tomo)
+
+    dz, dy, dx = cp.shape(padded_tomo)
+
+    # 3D FFT of tomo data
+    padded_tomo = cp.asarray(padded_tomo, dtype=cp.complex64)
```
dkazanc, line 524: any reason why it it converted to the complex array? the padded tomo should be float32, then the fft will result in complex64?

dkazanc: I guess this is pre-allocation of the memory. Should be fine... You can also do this in the beginning if you going to do padding using cupy? 
`tomo = cp.asarray(tomo, dtype=cp.float32)` and then do the padding afterwards

namannimmo10:
Here is the benchmark test that I use on tomo standard data:
```python
@cp.testing.gpu
@pytest.mark.perf
def test_paganin_filter2_performance(ensure_clean_memory):
    # Note: low/high and size values taken from sample2_medium.yaml real run

    # this test needs ~20GB of memory with 1801 - we'll divide depending on GPU memory
    dev = cp.cuda.Device()
    mem_80percent = 0.8 * dev.mem_info[0]
    size = 1801
    required_mem = 20 * 1024 * 1024 * 1024
    if mem_80percent < required_mem:
        size = int(np.ceil(size / required_mem * mem_80percent))
        print(f"Using smaller size of ({size}, 5, 2560) due to memory restrictions")

    data_host = np.random.random_sample(size=(size, 5, 2560)).astype(np.float32) * 2.0
    data = cp.asarray(data_host, dtype=np.float32)

    # run code and time it
    # cold run first
    paganin_filter2(data)
    dev = cp.cuda.Device()
    dev.synchronize()

    start = time.perf_counter_ns()
    nvtx.RangePush("Core")
    for _ in range(10):
        paganin_filter2(data)

    nvtx.RangePop()
    dev.synchronize()
    duration_ms = float(time.perf_counter_ns() - start) * 1e-6 / 10

    assert "performance in ms" == duration_ms
```
--------------------------------------------------------------------------------------------

| Stage | WS381  | Diamond cluster node |
| ------------- | ------------- | ----------|
| Before optimisation  | 895.359 ms  | 845.014 ms |
| After optimisation  | 12.1399 ms | 9.55 ms |

PR #114, opened 2024-01-17
Add initial implementation of the FBP algorithm with CIL

paskino:
Uses FBP recon with CIL and ASTRA backend for 2D and 3D parallel beam data using these parameters:
```
Parameters
    ----------
    data : np.ndarray
        Projection data as a NumPy array.
    angles : np.ndarray
        An array of angles given in radians.
    center : float, optional
        The center of rotation (CoR).
    objsize : int, optional
        The size in pixels of the reconstructed object.
    gpu_id : int, optional
        A GPU device index to perform operation on.
    filter : str, optional
        The Filter of the FBP algorithm, default 'ram-lak'
    num_slices : int, optional
        The number of slices to process in each chunk.
```
Requires to make sure the following are correct or correctly defined:

- [ ] definition of `center`: from the beginning or middle of pixel
- [ ] rotation angle direction


dkazanc:
Will merge it to the local branch `cil` and work on it.

PR #116, opened 2024-01-29
set the correct dimension labels

paskino:
Default `AcquisitionGeometry` expects [different dimension labels](https://github.com/TomographicImaging/CIL/blob/f9102f996ba9bfe89940c1ef1b67dfe9cca5de3e/Wrappers/Python/cil/framework/framework.py#L4234) than what is passed, leading to a

```
ValueError: Shape mismatch got (128, 180, 160) expected (180, 128, 160)
```

### DIFF
--- a/httomolibgpu/recon/algorithm.py
+++ b/httomolibgpu/recon/algorithm.py
@@ -146,6 +146,7 @@ def FBP_CIL(
             rotation_axis_position=(center - panel_centre, 0, 0)
         )
         ag.set_panel(num_pixels=(det_x, det_y), pixel_size=(1.0, 1.0))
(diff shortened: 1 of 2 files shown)


dkazanc:
thanks, still a  couple of issues there: 

- `astra.set_gpu_index(gpu_id)` results in _AttributeError: module 'collections' has no attribute 'Iterable'_. We haven't used this command before, I think we set it as `cfg["option"] = {"GPUindex": .device_index}`
- `reconstruction = fbp.run(verbose=0)` fails with _ValueError: Cannot fill with the provided array.Expecting shape (128, 180, 160) got (128, 128, 160)_, so may be something still off with the geometry?

dkazanc:
Just a thought as I swap the axes for the input data, you might need C-contiguous array?

Issue #57, opened 2023-02-07
Set up a pre-commit hook


namannimmo10:
Hi team, 
let me know if you are ok with me setting up a pre-commit hook: https://pre-commit.com/
We can add pre-commit plugins to this repo (like trailing whitespace, yaml checker, etc) with `.pre-commit-config.yaml` file.

dkazanc:
totally OK with that, thanks. Would it also suggest changes or just give you a report with issues?

PR #132, opened 2024-06-07
adding precommit hook

dkazanc:
#57


dkazanc:
image

namannimmo10:
Many projects prefer ruff these days: https://github.com/astral-sh/ruff-pre-commit.

dkazanc:
Will add it with `black` for now. For smaller projects it doesn't [seem](https://www.packetcoders.io/whats-the-difference-black-vs-ruff/) to be a huge difference.

PR #136, opened 2024-07-08
DOC: Customise `__repr__` for a better representation

namannimmo10:


namannimmo10:
CI is currently failing due to a different reason, let's fix it in a separate PR.

PR #140, opened 2024-07-18
CI: Add code coverage

namannimmo10:


.github/workflows/tests_run.yml
```diff
@@ -35,11 +35,21 @@ jobs:
...
+          coverage run -m pytest tests/
+          coverage xml
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          name: Code Coverage
+          fail_ci_if_error: true
+          files: ./coverage.xml
```
namannimmo10:
```suggestion
      - name: Upload coverage to Codecov
        env:
          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
        run: |
          bash <(curl -s https://codecov.io/bash) -f coverage.xml -F unittests
```

namannimmo10:
Unable to push coverage to Codecov as git is not available with this container image.

yousefmoazzam:
> Unable to push coverage to Codecov as git is not available with this container image.

Ah I see, that's a pain... Might be a dumb question, but is it possible to do an `apt install -y git` step (or something similar depending on what package manager is available in the container image) in the workflow to make git available? Not ideal, but if it works, that could be a workaround for now.

namannimmo10:
No, cannot have `sudo` permissions easily. Trying to find a workaround!

namannimmo10:
One option is to add a Dockerfile.

yousefmoazzam:
Eh, that's kinda annoying, not allowing use of `sudo` within a container - what harm can it do, it's a container...

But fair enough, good luck trying to find a workaround, feels like it might be a pain :sweat_smile:

yousefmoazzam:
FYI, there's a github action for uploading coverage to codecov, could be useful instead of trying to manually do it in a shell in a step in the workflow https://github.com/codecov/codecov-action (apologies if this has already been looked at and was found to be incompatible with the IRIS runner or something!)

namannimmo10:
Yeah - I tried https://github.com/DiamondLightSource/httomolibgpu/pull/140/commits/a7abd29e05374596f2210c33c538fbe316d0aaf9 before, but it's unable to find `git`. Think we are nearly there.

namannimmo10:
So peculiar. Same error again. Raised https://github.com/codecov/codecov-action/issues/1546.

yousefmoazzam:
I was just perusing the issues and PRs, looked at the issue raised with the codecov action, and this error repeated twice at the end got me thinking a bit:
```
POST: 404 - {'detail': ErrorDetail(string='Could not find a repository associated with upload token ***', code='not_found')}
```

Does the repo in the relevant github organisation (`DiamondLightSource`) for which we want to get code coverage need to be setup on https://app.codecov.io/github/DiamondLightSource?

If I search for "httomo", I can see that httomolibgpu has the "Configure" button on the right, which suggests that it hasn't been "configured" to have code coverage uploaded to it?
codecov-io-httomo

If that was the issue, I could maybe understand the error being returned; it's essentially saying that coverage is trying to be uploaded for a github repo that isn't configured for that in codecov's website.

The docs for codecov suggest that this needs to be done https://docs.codecov.com/docs/quick-start#step-1-sign-up-for-codecov

namannimmo10:
Hi, thanks for pointing this out, see this discussion: https://diamondlightsource.slack.com/archives/CKW8E0V4H/p1721394261955809.

yousefmoazzam:
Fair enough. Though in my opinion, because the responder said that they're "not sure", I would not consider this a proof that stuff on codecov doesn't need to be done. Especially as there's no explanation for _how_ it automatically gets done from simply being a repo in the DiamondLightSource organisation, and the fact that the codecov docs indicate that something does need to be done.

If I choose random repos from the DiamondLightSource organisation which are marked as "Configured" in codecov, they all are using the codecov github action, and if I choose random repos which are marked as "Not configured" in codecov, they do not use the codecov github action.

namannimmo10:
Hi Tom 👋 Many thanks for releasing `v5`. I just gave it a try and am still getting an upload error,

```console
==> Running upload-coverage
      ./codecov --verbose upload-coverage -t <redacted> --fail-on-error --git-service github --sha eea3f1d1cf4ff501da1fc8934212e2363d80100e --file /home/runner/coverage.xml --gcov-executable gcov --name httomolibgpu-codecov
[PYI-970:ERROR] Failed to load Python shared library '/tmp/_MEIUJQIIM/libpython3.11.so.1.0': dlopen: /usr/lib64/libm.so.6: version `GLIBC_2.29' not found (required by /tmp/_MEIUJQIIM/libpython3.11.so.1.0)
==> Failed to upload coverage
    Exiting...
```

Perhaps the custom container we are using here (`ghcr.io/diamondlightsource/httomolibgpu:dockerfile`) needs to have `GLIBC_2.29`, but again, I find it peculiar.

namannimmo10:
OK, I don't see an error with that now. The branch is set to `null`. From Code tree at https://app.codecov.io/github/DiamondLightSource/httomolibgpu:

> Once merged to your default branch, Codecov will show your report results on this dashboard

PR #145, opened 2024-08-14
Move import of RecToolsDIRCuPy and RecToolsIRCuPy to module scope

ptim0626:
This PR moves import of `RecToolsDIRCuPy` and `RecToolsIRCuPy` into module scope, instead of function scope.

Not only this is a more Pythonic style, but most importantly, it avoids duplicated references of input `data` and the output reconstruction. Importing `RecToolsDIRCuPy` and `RecToolsIRCuPy` inside the function seems creating unnecessary references of their CuPy arrays when the functions return to the caller. By importing at the module scope, this duplication of references is not present and ensure proper clean-up by the CuPy memory pool upstream.

### DIFF
--- a/httomolibgpu/recon/algorithm.py
+++ b/httomolibgpu/recon/algorithm.py
@@ -29,6 +29,9 @@ nvtx = cupywrapper.nvtx
 from numpy import float32, complex64
 from typing import Optional, Type
 
+from tomobar.methodsDIR_CuPy import RecToolsDIRCuPy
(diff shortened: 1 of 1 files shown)


dkazanc:
Thank a lot @ptim0626 , as I remember correctly why I moved the imports inside the function is because of some issues with the doc builds. I guess we could either mock `tomobar` or have it as a dependency for the docs build. I'll investigate it. Happy to merge it.

PR #146, opened 2024-08-15
Add Dockerfile and corresponding workflow

namannimmo10:
This is an attempt to build and push a docker image to the GitHub container registry using `nvidia/cuda:11.6.2-devel-ubi8` as a base image. This newly built image can then be used for our GA workflows.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,45 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    branches:
+      - "dockerfile"
+  workflow_dispatch:
+
+jobs:
+  build-and-push-image:
+    if: ${{ github.repository == 'DiamondLightSource/httomolibgpu' }}
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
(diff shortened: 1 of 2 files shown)


namannimmo10:
OK, the workflow succeeded. From the logs: 

> pushing ghcr.io/diamondlightsource/httomolibgpu:dockerfile with docker

yousefmoazzam:
Cool! :tada: Can indeed see the built container image in https://github.com/diamondlightsource/httomolibgpu/pkgs/container/httomolibgpu

dkazanc:
thanks @namannimmo10 , so currently nothing is running in the container but it has got the conda environment for httomolibgpu?

namannimmo10:
No this doesn't contain the dependencies for `httomolibgpu`. Hopefully, this will fix the issue we're having with #140 where `git` is unavailable. But similar approach could be followed to create a docker image for the library itself.

namannimmo10:
OK, this PR serves the purpose, I will merge it. We can certainly create an image for the library and push it with a different tag. This workflow won't run on the main branch.

PR #157, opened 2024-09-20
Rescale to int cpu/gpu agnostic

dkazanc:
CPU-GPU `rescale_to_int` function. All tests modified to compare the results of CPU with GPU results. 

I discovered that the GPU function accepts only float32 array as an input yet the CPU one can accept any data type (because of the cast to float64). The GPU implementation is significantly more memory efficient because of the elementwise loop and casting scalars instead of arrays compared to Python one-liners. 

So far I don't see the reason to satisfy other data type inputs, however, in future we might work with `uint16` then this code must be adapted accordingly. 

I'll modify image saver in [here](https://github.com/DiamondLightSource/httomolib/pull/17) accordingly, we need to remove redundant parameters.


httomolibgpu/misc/rescale.py
```diff
@@ -76,13 +77,18 @@ def rescale_to_int(
     else:
         output_dtype = np.uint32
 
+    if cupy_run:
+        xp = cp.get_array_module(data)
+    else:
+        import numpy as xp
+
```
yousefmoazzam, line 87: Given that [`cp.get_array_module()`](https://docs.cupy.dev/en/v12.3.0/reference/generated/cupy.get_array_module.html) is intended to return either the cupy or numpy module depending on if any of the given args require cupy, I would have thought that doing simply
```python
xp = cp.get_array_module(data)
```

would work.

If I do that locally and run the tests they all pass, which suggests that it does work without the if/else.

I do understand though that the CPU/GPU agnostic stuff can be tricky and that there may be other factors at play here, so I wanted to ask: is there some reason I'm unaware of that the if/else is needed?

yousefmoazzam:
Apologies for the broken IRIS CI job, it looks like maybe IRIS isn't happy, the conda env failed to be created... https://github.com/DiamondLightSource/httomolibgpu/actions/runs/11052312862/job/30704122840?pr=157

httomolibgpu/misc/rescale.py
dkazanc: this is for the true GPU-less case, so then  `cp.get_array_module(data)` won't be possible without the condition. This makes sure that this can be run on the CPU arch alone.

dkazanc:
Detailed docs page added
https://diamondlightsource.github.io/httomolibgpu/reference/methods_list/rescale_to_int.html

Seems like IRIS is down, I'm merging this PR...

httomolibgpu/misc/rescale.py
yousefmoazzam: Ok, I see.

I know this PR is merged already, but I figured it might be worth raising this point anyway. This feels like it's kinda pointless to be using `cp.get_array_module()` then. `get_array_module()` is intended to be a generic way to get either the `cupy` or `numpy` module, but in our case this generic way doesn't work if the user has no GPU.

As the code is currently, it's saying: "if we have a GPU, then get the module generically with `get_array_module()`, and if we don't have a GPU then use `numpy`". If someone has a GPU, then what's the point of getting the module "generically" with `get_array_module()` - they have a GPU, so why not just `import cupy as xp`?

It feels like it'd be clearer to just do:
```python
    if cupy_run:
        import cupy as xp
    else:
        import numpy as xp
```

because I don't really see how using `get_array_module()` is doing anything useful, and it makes the condition harder to understand.

dkazanc: I see your point...  it is a bit redundant as we know that there is a GPU and also CuPy available. I'll do the change.

Issue #160, opened 2024-09-27
Versioning system


dkazanc:
I guess it also applies to other HTTomo packages so we can decide more generally may be @yousefmoazzam ?
When we create a release we trigger the CI job that build this package labelled as `main`.  At midnight, however, we have the auto build that creates a `dev` build using the same tag. So the original `main` release gets [replaced](https://anaconda.org/httomo/httomolibgpu/files)by the `dev` release. 

To fight this problem I usually create 2 `main` releases, the one for the current release and one for the future release which will become the `dev` release at midnight, hence keeping the previous `main` release intact. Then, when I ready to do another release I will delete the tag associated with the `dev` release and re-introduce it again therefore making the `dev` release `main` instead. 

I do appreciate that this might be slightly confusing and it probably should be written somewhere in the repo supporting info. But, in general ,do we agree that this is something we want to do, keep both `dev` and `main` ? Or, just stick to `main` for instance and drop `dev` entirely? I'm not entirely sure that `dev` will be used that much, at least for the libraries.  Then also with PyPy in place we probably be working more with pip installed libraries, so we can keep just one conda main package?

yousefmoazzam:
If I remember correctly, the idea behind the nightly build and uploading to the `dev` label was to be able to test out the latest code in the package from an installation of a conda package, rather than a developer using their local git repo.

If this isn't a common use case for any of the developers (personally, I haven't installed any of the nightly builds of the conda package yet), then it might make sense to stop doing the nightly builds, which would then mean we have no need to upload packages to the `dev` label.

dkazanc:
cheers, I think testing the `dev` package certainly makes sense. This is certainly needed  for more complex and involved software, e.g., `httomo`. I don't think that this is that needed for the libraries as the development itself is quite linear most of the time? It will also probably make our life slightly easier by just releasing the `main` package? 

But if we want to keep doing `dev` releases, then we also need to include the CI job of testing them, otherwise, indeed it is a bit useless.

yousefmoazzam:
For the point about the backend packages like httomolib and httomolibgpu probably not needing `dev`, yeah, I'd agree that uploading packages to the `main` label is enough.

For the point about "testing `dev`", I realise maybe I should have chosen my words more carefully. Before I said this:

> If I remember correctly, the idea behind the nightly build and uploading to the `dev` label was to be able to test out the latest code in the package from an installation of a conda package, rather than a developer using their local git repo.

[...]

PR #161, opened 2024-10-02
distortion correction preview refactoring

dkazanc:
fixes #157 #160  and #163


yousefmoazzam (commented):
Changes to the distortion correction method looks good to me :slightly_smiling_face:  Though I don't understand why there's a random change to the nightly dev workflow file in this PR?

It's a bit muddled to have two totally unrelated issues tackled in the same PR, but if that must be done, could the "backup" file be removed completely instead of just being renamed? The backup of the file will exist in past states of the repo via version control, so removal of the file is cheap to do (ie, it's easy to find the backup file), without needing to keep the backup file under version control.

httomolibgpu/prep/alignment.py
```diff
@@ -90,12 +91,10 @@ def distortion_correction_proj_discorpy(
...
-    shift = preview["starts"]
-    step = preview["steps"]
-    x_dim = 1
-    y_dim = 0
-    step_check = max([step[i] for i in [x_dim, y_dim]]) > 1
-    if step_check:
+    det_x_step = step[0]
+    det_y_step = step[1]
+
+    if det_y_step > 1 or det_x_step > 1:
```
yousefmoazzam, line 97: I just noticed that the error message should also be updated to not mention previewing

dkazanc:
OK, I got rid of the `dev` CI backup and also changed tuples for `shift` and `step` into lists. I believe there is an issue to cast a tuple correctly in httomo resulting in a something like this after yaml generator:
```
    pad: !!python/tuple
    - 0
    - 0
```
There is quite a lot of problems like this with Tomopy functions.
Feel free to [add `preview` string to parameters name](https://github.com/DiamondLightSource/httomo/issues/455#issuecomment-2399409524) if you think it will be a better way to search for them in  `httomo`.

yousefmoazzam:
Oh right, with the mention of the parsing YAML into a python tuple, are we planning on having users entering the values for `shift` and `step` in the YAML directly?

Meaning, are we going to have something like the following in the YAML for the distortion correction:
```yaml
- method: standard_tomo
  module_path: httomo.data.hdf.loaders
  parameters:
    name: tomo
    data_path: entry1/tomo_entry/data/data
    image_key_path: entry1/tomo_entry/instrument/detector/image_key
    rotation_angles:
      data_path: /entry1/tomo_entry/data/rotation_angle
  preview:
    detector_y:
      start: DET_Y_START
      step: DET_Y_STEP
    detector_x:
      start: DET_X_START
      step: DET_X_STEP
.
.
.
- method: distortion_correction_proj_discorpy
  module_path: httomolibgpu.prep.alignment
  parameters:
    metadata_path: /some/path
    shift: [DET_X_START, DET_Y_START]
    step: [DET_X_STEP, DET_Y_STEP]
    order: 1
    mode: reflect
```

where the users have some value for `DET_Y_START`, `DET_X_START`, `DET_Y_STEP`, and `DET_X_STEP` that they put in two places?

The reason I ask is because I _think_ the YAML would need to be parsed into a tuple only if the YAML contains the actual values to go into the distortion correction method?

But then in order for the values for `shift` and `step` to be put into the YAML (to be parsed into a python tuple), the values need to be duplicated in the YAML config:
- once in the loader's `preview`
- once in the distortion correction's `shift` and `step`

Or have I misunderstood how this would work (which is very possible at this point...)?

dkazanc:
I see, we're indeed would like to deal with those parameters behind the scenes in HTTomo, so there is no need for us to expose those parameters in YAML to users? Yes, it makes sense then. We've got `glob_stats` as a tuple as well, so I don't mind tuple or list, as it will be hidden anyway.

httomolibgpu/prep/alignment.py
```diff
@@ -48,7 +48,8 @@
 def distortion_correction_proj_discorpy(
     data: cp.ndarray,
     metadata_path: str,
-    preview: Dict[str, List[int]],
+    shift: List[Union[int, int]] = [0, 0],
+    step: List[Union[int, int]] = [1, 1],
```
yousefmoazzam: If we go with a list, I think the correct type would be `List[int]`. Currently, the type `List[Union[int, int]]` contains the redundant type `Union[int, int]` which means "an `int` or an `int`". Altogether, the type `List[Union[int, int]]` means "a list of values with type `int` or `int`", which I imagine is not what was intended, and the intention more likely was "a list of values of type `int`".

If we go with a tuple, the type should be `Tuple[int, int]`.

httomolibgpu/prep/alignment.py
```diff
@@ -63,11 +64,11 @@ def distortion_correction_proj_discorpy(
...
-    preview : Dict[str, List[int]]
-        A dict containing three key-value pairs:
-        - a list containing the `start` value of each dimension
-        - a list containing the `stop` value of each dimension
-        - a list containing the `step` value of each dimension
+    shift: List, optional
+         Centers of distortion in x (from the left of the image) and y directions (from the top of the image).
+
+    step: List, optional
+         Steps in x and y directions respectively. They need to be not larger than one.
```
yousefmoazzam: Similar to the other comment, a note about types.

If we go with a list, the type should be `List[int]` (no `optional`).

If we go with a tuple, the type should be `Tuple[int, int]` (no `optional`)

httomolibgpu/prep/alignment.py
```diff
@@ -33,7 +33,7 @@
 else:
     map_coordinates = Mock()
 
-from typing import Dict, List
+from typing import Dict, List, Tuple, Union
```
yousefmoazzam: Two things to note:
1. `Union` isn't needed, regardless of if we choose a list or tuple for the `shift` and `step` params
2. remove `Tuple` import if we go with a list

yousefmoazzam (commented):
Due to it still not being clear how we'll change parameters in the method to allow httomo to pass the relevant info from the loader's `preview` to the method, I won't make any changes here yet, just some comments about types.

PR #167, opened 2024-10-30
Raven filter

neon60:
* Add Raven filter to strip.py
* Add Raven filter unit tests
* Add Raven filter performance tests


docs/source/examples/raven_filter_example.py
```diff
@@ -0,0 +1,86 @@
+import os
+import numpy as np
+import cupy as cp
+import scipy
+import pyfftw
```
namannimmo10: `pyfftw` can be added as a developer dependency since it is used in the reference CPU implementation.

tests/test_prep/test_stripe.py
```diff
@@ -3,14 +3,63 @@
...
 from httomolibgpu.prep.normalize import normalize
 from httomolibgpu.prep.stripe import (
     remove_stripe_based_sorting,
     remove_stripe_ti,
     remove_all_stripe,
+    raven_filter,
 )
 from numpy.testing import assert_allclose
 
+def raven_filter_cpu(
```
namannimmo10: Shall we add this to a separate file and import it? 

See https://github.com/DiamondLightSource/httomolibgpu/tree/main/tests/test_recon.

tests/test_prep/test_stripe.py
```diff
@@ -66,6 +114,20 @@ def test_stripe_removal_sorting_cupy(data, flats, darks):
     assert corrected_data.dtype == np.float32
     assert corrected_data.flags.c_contiguous
 
+def test_stripe_raven_cupy(data, flats, darks):
+    # --- testing the CuPy port of TomoPy's implementation ---#
+
+    data = normalize(data, flats, darks, cutoff=10, minus_log=True)
+
+    data_after_raven_gpu = raven_filter(cp.copy(data)).get()
+    data_after_raven_cpu = raven_filter_cpu(cp.copy(data))
```
namannimmo10: We can use [`numpy_cupy_allclose`](https://docs.cupy.dev/en/stable/reference/generated/cupy.testing.numpy_cupy_allclose.html). See 

https://github.com/DiamondLightSource/httomolibgpu/blob/a77f3346460d26a9c490cbb0675df6ed2ea45266/tests/test_recon/test_rotation.py#L142

docs/source/examples/raven_filter_example.py
neon60: Fixed.

tests/test_prep/test_stripe.py
neon60: Fixed.

tests/test_prep/test_stripe.py
neon60: Fixed.

namannimmo10: Sorry I meant to say that we should be using `xp` to test both the CPU and GPU methods. We can use a random generator - see how the following test is written, for instance:

```python
@pytest.mark.parametrize("norm", [False, True], ids=["no_normalise", "normalise"])
@pytest.mark.parametrize("overlap", [False, True], ids=["no_overlap", "overlap"])
@pytest.mark.parametrize("denoise", [False, True], ids=["no_denoise", "denoise"])
@pytest.mark.parametrize("side", [1, 0])
@cp.testing.numpy_cupy_allclose(rtol=1e-5, atol=1e-6)
def test_find_center_360_unity(ensure_clean_memory, xp, norm, overlap, denoise, side):
    # because it's random, we explicitly seed and use numpy only, to match the data
    np.random.seed(12345)
    data = np.random.random_sample(size=(128, 1, 512)).astype(np.float32) * 2.0 + 0.001
    data = xp.asarray(data)

    if xp.__name__ == "numpy":
        (cor, overlap, side, overlap_pos) = find_center_360_numpy(
            data, use_overlap=overlap, norm=norm, denoise=denoise, side=side
        )
    else:
        (cor, overlap, side, overlap_pos) = find_center_360(
            data, use_overlap=overlap, norm=norm, denoise=denoise, side=side
        )

    return xp.asarray([cor, overlap, side, overlap_pos])
```

If the results don't match, it will fail.

tests/test_prep/test_stripe.py
```diff
@@ -116,6 +131,48 @@ def test_remove_stripe_ti_performance(ensure_clean_memory):
...
+    for _ in range(10):
+        # have to take copy, as data is modified in-place
+        raven_filter(cp.copy(data))
+    nvtx.RangePop()
+    dev.synchronize()
+    duration_ms = float(time.perf_counter_ns() - start) * 1e-6 / 10
+
+    assert "performance in ms" == duration_ms
+
+# @pytest.mark.perf
```
namannimmo10: We can remove it instead of commenting it out.

httomolibgpu/prep/stripe.py
```diff
@@ -359,6 +366,96 @@ def _rs_dead(sinogram, snr, size, matindex, norm=True):
...
+
+    Raises
+    ------
+    ValueError
+        If the input array is not three dimensional.
+    """
+
+    input_type = sinogram.dtype
+
+    if input_type not in ["float32"]:
```
namannimmo10:
```suggestion
    if sinogram.dtype != cp.float32:
```

tests/test_prep/test_stripe.py
neon60: Added a parameters test separately, where ```xp``` is used.

tests/test_prep/test_stripe.py
neon60: Removed.

dkazanc:
thanks @neon60. I've also added you to the contributors list so you can create branches from `httomolibgpu` directly. In that case the CI of ours that run tests should work (I suspect that it is a reason for the current failure). I will test the method on a real data.

namannimmo10 (approved):
Looks good, thank you Istvan!

PR #169, opened 2024-12-11
Vocentering fixes

dkazanc:
making changes to replicate the Savu's implementation as well as adding two additional parameters: average_radius and cor_initialisation_value.

Optimised CUDA kernels are currently switched off.


dkazanc:
tests actually pass for me locally... something weird with iris CI again.

PR #177, opened 2024-12-20
changing the docstrings

itsdiy0:
changed the docstring of misc.morph.data_resampler and recon.rotation.find_center_pc methods to matches the pattern of other methods docstrings

### DIFF
--- a/httomolibgpu/misc/morph.py
+++ b/httomolibgpu/misc/morph.py
@@ -126,7 +126,7 @@ def data_resampler(
         ValueError: When data is not 3D
 
     Returns
-    ----------
+    -------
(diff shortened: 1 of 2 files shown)


httomolibgpu/misc/morph.py
```diff
@@ -104,21 +104,29 @@ def sino_360_to_180(
...
+    interpolation : str, optional
+        Selection of interpolation method. Defaults to 'linear'.
+
+    Raises
+    ----------
         ValueError: When data is not 3D
 
-    Returns:
+    Returns
+    ----------
```
namannimmo10:
```suggestion
    Returns
    -------
```

PR #184, opened 2025-01-11
Vocentering gpu offload

neon60:
**Improve performance of vo search**

- Use full mask
- Use modified _calculate_metrics

I will validate these again:
- The following CPU implementation has no significant effect on performance, that's why I did not port to GPU: https://github.com/DiamondLightSource/httomolibgpu/blob/vocentering_gpu_offload/httomolibgpu/recon/rotation.py#L191-L202
- The downsample kernel also did not give significance improvement.

Please ignore these files, I will delete them:
- [docs/source/metrics_calculate.py](https://github.com/DiamondLightSource/httomolibgpu/pull/184/files#diff-3a623bd4c538a183b379f10c9ba2f6e322cd320acb9854c3fea51dae61248b52)
- [docs/source/rotation_new.py](https://github.com/DiamondLightSource/httomolibgpu/pull/184/files#diff-0cf2d45f9a1f58a8e8d23239f1d1637a874092a61c2cb704171b27a89a493a58)


dkazanc:
thanks @neon60 I'm going through it now. We added real data tests into the library recently so I'll add that. Also will clean some comments as I go.

namannimmo10:
You must add a label to run the Zenodo tests - with https://github.com/DiamondLightSource/httomolibgpu/pull/184/commits/6ae3cb16c02fe67b18cad35d769357278150f67e the big data tests didn't run. The tests will run only after you add the label and make a commit.

namannimmo10:
Currently the CI fails because the performance test was collected 

> FAILED zenodo-tests/test_recon/test_rotation.py::test_center_vo_i12_dataset1_performance

PR #185, opened 2025-01-16
adding more zenodo tests and update default FBP parameters

dkazanc:


dkazanc:
cupy error again, zenodo tests pass locally. merging

PR #199, opened 2025-02-26
Improve the performance of stripe removal based on sorting

namannimmo10:
Applies sorting directly on the sinogram along the second axis. There's no need to store and manipulate sorting indices, hence reducing memory overhead.

Performance comparison done on ws582:

Using https://github.com/DiamondLightSource/httomolibgpu/commit/c83f021128e74f2a0ea831955b35da719b3a43f3 

```console
AssertionError: assert 'performance in ms' == 66.43685479999999
```

Using `main`

```console
AssertionError: assert 'performance in ms' == 193.274476
```

**~3x speedup**
```
>>> 193.274476 / 66.43685479999999
2.909145482305403
>>>
```

### DIFF
--- a/httomolibgpu/prep/stripe.py
+++ b/httomolibgpu/prep/stripe.py
@@ -93,24 +93,17 @@ def remove_stripe_based_sorting(
 
 
 def _rs_sort(sinogram, size, dim):
-    """
-    Remove stripes using the sorting technique.
-    """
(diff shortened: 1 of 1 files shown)


namannimmo10:
This change isn't correct, closing ...

namannimmo10:
The tests aren't robust enough to catch the error!

PR #200, opened 2025-02-27
Zenodo stripes removal tests

dkazanc:
Adding stripe removal tests for a new data from zenodo that contains stripes.

### DIFF
--- a/zenodo-tests/test_prep/test_stripe.py
+++ b/zenodo-tests/test_prep/test_stripe.py
@@ -129,7 +129,7 @@ def test_remove_all_stripe_i12_dataset4(
     norm_res = np.linalg.norm(residual_calc.get().flatten())
 
     assert isclose(norm_res, norm_res_expected, abs_tol=10**-4)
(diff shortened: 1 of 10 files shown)


zenodo-tests/test_prep/test_stripe.py
```diff
@@ -0,0 +1,176 @@
...
+        sm_size=sm_size_val,
+        dim=1,
+    )
+
+    residual_calc = data_normalised - output
+    norm_res = np.linalg.norm(residual_calc.get().flatten())
+
+    assert isclose(norm_res, norm_res_expected, abs_tol=10**-4)
+
+    assert output.dtype == np.float32
```
namannimmo10: All looks good. Just an extra sanity check!
```suggestion
    assert isclose(norm_res, norm_res_expected, abs_tol=10**-4)
    assert not np.isnan(output).any(), "Output contains NaN values"
    assert output.dtype == np.float32
```

Issue #212, opened 2025-05-21
stripe_remove_ti method data size change


dkazanc:
`stripe_remove_ti` method returns data with different dimension. 
Based on even or odd number of projections in the input, `stripe_remove_ti` method can change the size of the output volume.

PR #214, opened 2025-06-02
Data validation checker

dkazanc:
adding data validation modules, tests and correcting all functions in misc

### DIFF
--- a/httomolibgpu/misc/supp_func.py
+++ b/httomolibgpu/misc/supp_func.py
@@ -102,7 +102,7 @@ def _naninfs_check(
     if present_nans_infs_b:
         if verbosity:
             print(
-                f"Warning!!! Input data to method: {method_name} contains Inf's or/and NaN's. This will be corrected but it sometimes recommended to check the validity of input to the method."
+                f"Warning!!! Input data to method: {method_name} contains Inf's or/and NaN's. This will be corrected but it is recommended to check the validity of input to the method."
             )
 
     return data
--- a/httomolibgpu/misc/supp_func.py
+++ b/httomolibgpu/misc/supp_func.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# ---------------------------------------------------------------------------
+# Copyright 2022 Diamond Light Source Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ---------------------------------------------------------------------------
+# Created By  : Tomography Team at DLS <scientificsoftware@diamond.ac.uk>
+# Created Date: 02/June/2025
+# ---------------------------------------------------------------------------
+"""This is a collection of supplementary functions (utils) to perform various data checks"""
+
+from httomolibgpu import cupywrapper
+from typing import Optional
+
+cp = cupywrapper.cp
+cupy_run = cupywrapper.cupy_run
+
+
+def _naninfs_check(
+    data: cp.ndarray,
+    correction: bool = True,
+    verbosity: bool = True,
+    method_name: Optional[str] = None,
+) -> cp.ndarray:
+    """
+    Function finds NaN's, +-Inf's in the input data and then prints the warning and correct the data
+
+    Parameters
+    ----------
+    data : cp.ndarray
+        Input CuPy array either float32 or uint16 data type.
+    correction : bool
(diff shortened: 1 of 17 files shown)


tests/test_misc/test_supp_func.py
```diff
@@ -0,0 +1,190 @@
...
+import numpy as np
+
+from httomolibgpu.misc.supp_func import (
+    _naninfs_check,
+    _zeros_check,
+    data_checker,
+)
+from numpy.testing import assert_allclose, assert_equal
+
+eps = 1e-6
```
yousefmoazzam: Is `eps` used anywhere, I can't seem to find any other references to it in the file?

tests/test_prep/test_stripe.py
```diff
@@ -34,25 +34,13 @@ def test_remove_stripe_ti_on_data(data, flats, darks):
...
-#     assert_allclose(
-#         np.median(corrected_data), np.median(corrected_host_data), rtol=1e-6
-#     )
+# @pytest.mark.parametrize("angles", [180, 181])
+# @pytest.mark.parametrize("det_x", [11, 18])
+# @pytest.mark.parametrize("det_y", [5, 7])
+# def test_remove_stripe_ti_dims_change(angles, det_y, det_x):
+#     data = cp.random.random_sample(size=(angles, det_y, det_x)).astype(cp.float32) * 2.0
+#     corrected_data = remove_stripe_ti(data.copy())
+#     assert corrected_data.shape == (angles, det_y, det_x)
```
yousefmoazzam: Is this test intended to be commented out, or should it have been taken care of in some way?

httomolibgpu/misc/supp_func.py
```diff
@@ -0,0 +1,154 @@
...
+    method_name: Optional[str] = None,
+) -> bool:
+    """
+    Function that performs the variety of checks on input data, in some cases also correct the data and prints warnings.
+    Currently it checks for: the presence of infs and nans in data; the number of zero elements. 
+
+    Parameters
+    ----------
+    data : xp.ndarray
+        Input CuPy or Numpy array either float32 or uint16 data type.
```
yousefmoazzam, line 170: Out of curiosity, what happens if the data type is neither float32 or uint16, will the data be cast to some other type?

yousefmoazzam (commented):
Looks reasonable to me!

If it hasn't been done so already, I'd recommend running the memory hook tests in httomo-backends against these changes, to double-check that memory estimation of methods which use the new validation function hasn't been affected.

Another thing to mention is that httomolibgpu has a `.pre-commit-config.yaml` which has black formatting setup, so I'm assuming that httomolibgpu code should be formatted with black? If so, you may want to check if black has been applied to the changed files, I can see some things like missing newline characters and no spaces after commas that I _think_ wouldn't be present if black were applied (though I could be wrong!).

tests/test_prep/test_stripe.py
dkazanc: ah yeah, sorry, I was actually playing to see how this [issue](https://github.com/DiamondLightSource/httomolibgpu/issues/212) is affected. This test fails, I'll move that to another branch.

httomolibgpu/misc/supp_func.py
dkazanc: yes, I think the description is confusing here. It works with 64 bit data and probably will with 8-bit as well. I'll make a test to make sure. thx

dkazanc:
Very good point about `httomo-backends` tests, thanks. I see that it does affect them and some of the tests failing. I'll do some research, but it looks there are mainly two lines to blame? 
`xp.nan_to_num(data, copy=False, nan=0.0, posinf=0.0, neginf=0.0)`
and 
`zero_elements_total = int(xp.count_nonzero(data == 0))`

dkazanc:
This also allocates the memory:
`xp.all(xp.isfinite(data))`

dkazanc:
Yes, looks like both functions (inf's nan's estimation/removal) and zeros calculation create the copy of the data. 

This actually opens up more general discussion, sorry for many words. 

I can see two ways of dealing with this as modifying every memory estimator doesn't seem appealing at all. 
1. Write bespoke element-wise kernels for CuPy arrays to deal with the data without creating any copies of it. I reckon we will need to loop through arrays and fix OR not nan's/inf's. Same with zeros. It will be probably more computationally efficient as now. 
2. Have data validator as a separate method in the library with its own memory estimator and insert it using the framework between **each** method. Similar to what we do with `data_reducer` added after the loader by default. The data validator will be removed from the methods themselves in `httomolibgpu` library as it becomes a standalone method. 


Both approaches have benefits actually:
- no. 1 Seems like a nice algorithmic workaround and probably not very time consuming to do. Also I don't like that the fact that  such a simple method creates copies of data, that is probably not very efficient? 
- no. 2 Has a benefit of having this method for any other backend, it will also take care of httomolib and tomopy methods, for instance. Notably, we do not apply data validator to `data_reducer` and `save_to_images` of httomolib. In addition to that, it makes simpler for us to log warnings with the framework coming from the method. We can modify it in a way so that it will also return some info about the presence of nans/infs/zeros. 

To me it looks like we need both 1 and 2 :) But we can do this in stages, if no. 1 resolves the memory issue, we can merge it first and then we can do no. 2. Any thoughts?

dkazanc:
Latest update. I've implemented no.1 from above. All tests now pass in the branch and with `httomo-backends`.

PR #220, opened 2025-08-18
padding parameter for the horizontal detector (tomobar v. 2025.08)

dkazanc:
Changes to accommodate new API in  tomobar v. 2025.08 for a new parameter to enable edge-padding for the horizontal detector.

### DIFF
--- a/httomolibgpu/recon/algorithm.py
+++ b/httomolibgpu/recon/algorithm.py
@@ -59,6 +59,7 @@ def FBP2d_astra(
     data: np.ndarray,
     angles: np.ndarray,
     center: Optional[float] = None,
+    detector_pad: int = 0,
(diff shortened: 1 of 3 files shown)


yousefmoazzam (approved):
Looks good to me!

One thing I'd mention as a very minor point on docstrings is that there's frequent use of the phrase "horizontal detector width padding". This slightly confused me at first, thinking that the horizontal detector dimension has a "width", before understanding that it's referring to "detector width padding", or "horizontal detector dimension padding".

I think there's some redundancy in "horizontal detector width padding" and it'd maybe be clearer as either "detector width padding" or "horizontal detector dimension padding". But, I'm also fairly unfamiliar with tomobar's recon methods, so it could be that this is the usual way to refer to detector dimensions.

PR #221, opened 2025-08-21
Making rescale_to_int GPU only function

dkazanc:
Removing the CPU part from rescale_to_int function as there is a dedicated module in `httomolib` package

### DIFF
--- a/httomolibgpu/misc/rescale.py
+++ b/httomolibgpu/misc/rescale.py
@@ -30,7 +30,6 @@ from typing import Literal, Optional, Tuple, Union
 
 from httomolibgpu.misc.supp_func import data_checker
 
-from sofia.core.regularisers import rescale_to_int_C
 
 __all__ = [
     "rescale_to_int",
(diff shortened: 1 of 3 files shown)


httomolibgpu/misc/rescale.py
```diff
@@ -82,18 +83,13 @@ def rescale_to_int(
 
     data = data_checker(data, verbosity=True, method_name="rescale_to_int")
 
-    if cupy_run:
```
yousefmoazzam, line 85: Due to this deletion, it looks like the `cupy_run` variable is no longer used, and thus the import of it at the top of the module can be removed.

tests/test_misc/test_rescale.py
```diff
@@ -116,22 +95,14 @@ def test_rescale_double_offset_min_percentage(bits: Literal[8, 16, 32]):
...
 def test_tomo_data_scale(data):
     data_cpu = data.get()
     res_dev = rescale_to_int(
         data.astype(cp.float32), perc_range_min=10, perc_range_max=90, bits=8
     )
-    res_cpu = rescale_to_int(data_cpu, perc_range_min=10, perc_range_max=90, bits=8)
     res = res_dev.get()
     assert res_dev.dtype == np.uint8
-    assert res_dev.dtype == np.uint8
-    np.testing.assert_array_equal(res_cpu, res)
```
yousefmoazzam: After these deletions, `data_cpu` and `res` appear to not be referred to anymore, are they needed or can they also be deleted?

yousefmoazzam (commented):
CI is being weird I think, the tests pass locally for me :shrug: 

Only a few small things regarding unused variables and imports, otherwise looks good!

PR #245, opened 2026-01-05
Paganin filter: `padding_method` options

mfep:
Added `padding_method` parameter to `paganin_filter`. Options:
- `next_power_of_2` (default, original method): Pads data to the size that is the next power of 2
- `next_fast_length`: Pads data to the next size that is fast for FFT. This can be zero padding. Uses less memory than `next_power_of_2`
- Integer: specifies padding extent in pixels.

Updated tests and zenodo tests.

Fixes: [Paganin filter padding options](https://github.com/RadWay-Tech-Services/issue/issues/63)

### DIFF
--- a/httomolibgpu/prep/phase.py
+++ b/httomolibgpu/prep/phase.py
@@ -60,7 +60,7 @@ def paganin_filter(
     calculate_padding_value_method: Literal[
         "next_power_of_2", "next_fast_length", "use_pad_x_y"
     ] = "next_power_of_2",
-    pad_x_y: Optional[Tuple[int, int]] = None,
+    pad_x_y: Optional[list] = None,
     calc_peak_gpu_mem: bool = False,
 ) -> cp.ndarray:
     """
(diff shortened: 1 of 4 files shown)


httomolibgpu/prep/phase.py
```diff
@@ -56,6 +57,9 @@ def paganin_filter(
     distance: float = 1.0,
     energy: float = 53.0,
     ratio_delta_beta: float = 250,
+    padding_method: Union[
```
dkazanc: I think `padding_method` parameter can be slightly confusing to users. I would think that the padding method is something how you deal with the boundary, e.g. edge/symmetric/reflect etc. I believe we do not expose such option for now and by default it is always `edge` . Here it is how the padding value is calculated. So may be this parameter should be called `calculate_padding_value_method` instead? But then,  we also allow to provide an integer to it, which makes it slightly ambiguous. 
Also the padding integer should be a list,  separated for X and Y dimensions. So may be we should remove this option for now and provide only "next_power_of_2", "next_fast_length" ? Hope it makes sense?

httomolibgpu/prep/phase.py
```diff
@@ -26,6 +26,7 @@
 
 cp = cupywrapper.cp
 cupy_run = cupywrapper.cupy_run
+next_fast_len = cupywrapper.next_fast_len
```
dkazanc, line 29: I'd rather make an import from `scipy` here, `cupywrapper` is mostly to deal with the CuPy related imports

PR #265, opened 2026-09-02
normalize.py: Use `count_nonzero` in place of the reduction kernels

mfep:
Closes https://github.com/RadWay-Tech-Services/Diamond_Light_Source/issues/97


httomolibgpu/prep/normalize.py
```diff
@@ -180,9 +160,9 @@ def dark_flat_field_correction(
             50.0  # warning if more clipped values than given percentage
         )
 
-        clipped_total_up = int(count_greater_kernel(out, float32(upper_bound)))
+        clipped_total_up = cp.count_nonzero(out >= upper_bound)
```
dkazanc, line 163: I remember with CuPy 12.3 I started with this in-built function but it was performing poorly. I guess it has been improved. Will test.

